PR #1, opened 2018-01-16
Added docs using slate

cyrusaf:
Added docs using [slate docs](https://github.com/lord/slate). Added make commands to automatically setup/dev/build/publish docs to gh-pages. Slate requires ruby >= 2.3.1.

TODO:
* docs/README


hyandell:
Noting that the code of conduct should use our standard text/approach, rather than slate's. See oss-dashboard's CONTRIBUTING file ( https://github.com/amzn/oss-dashboard/ ).

docs/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2008-2013 Concur Technologies, Inc.
```
hyandell: This is a source header and not a license file.

Are Concur the author of Slate? ie: It's injecting its license into its output?

cyrusaf: I believe they are. I'm new to open source -- is it ok to remove the docs/LICENSE file? Almost everything inside docs/* is cloned from the slate github repo.

Makefile
```diff
@@ -30,6 +30,21 @@ setup:
...
+	bundler install --gemfile docs/Gemfile
+
+dev_docs: setup_docs
+	cd docs && bundle exec middleman server
+
+build_docs: setup_docs
+	cd docs && bundle exec middleman build --clean
+
+publish_docs: build_docs
+	git add docs/build && git push origin `git subtree split --prefix docs/build`:gh-pages --force
```
cyrusaf: I want to get 👀 on this command. It builds the html files, then automatically pushes them to the gh-pages branch of whichever fork/repo you are working on. Is there a better way of doing this (a way to review the outgoing changes to the docs/build on github before they are pushed to `gh-pages`)?

spenczar: The `docs/build` folder will just hold the built HTML documents, we can still review the source, right?

It would be nice if we could push a preview, yes. I don't know that GH pages has the flexibility to allow it, though. This might be the best we'll get within github pages.

spenczar:
Slate really is kind of a pain to work with. Let's follow github.com/golang/dep's lead and use http://docusaurus.io/. I like the way the docs look and I _especially_ like how simple it is.

PR #8, opened 2018-01-17
License fixes

hyandell:
Adding missing files; fixing license to be standard; adding license to readme

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,4 @@
+
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -178,15 +179,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "[]"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2018 Twitch Interactive, Inc.
+   Copyright [yyyy] [name of copyright owner]
```
nawuko: was this intended?

spenczar: good question for @hyandell

hyandell: Yes - that part of the license is an example source header rather than somewhere to put your actual copyright. By keeping it as the template it keeps the LICENSE file an easy match.

(with 20 years of hindsight, probably would have been better to have it on a webpage and not in the LICENSE file :) )

Issue #5, opened 2018-01-17
Use Continuous build/test of some kind


cep21:
Would using circle-ci or travis for the repository help?  It would allow people to submit pull requests and see the build result of them in the github UI.

spenczar:
Yes, we should definitely do this right away. Just haven't gotten around to it in the conversion from a private project to a public one.

hyandell:
Make sure you're checking the permissions they are asking for. Your GitHub org should have "Third-party application access policy=Access restricted". Travis CI (not Travis CI Private) is a good one.

spenczar:
Fixed in #25.

PR #25, opened 2018-01-18
Add travis

GarethLewin:
Adding travis config file (see https://travis-ci.org/twitchtv/twirp)

This is not complete, there are a few questions I would like to have answered.

First the readme says go 1.7+ but some of the helpers use `t.Helper()` which is a go1.9 feature. We can use compile pragmas to stop those tests from testing in 1.7 or we can change travis to only run in 1.9+

Second I didn't add linting because the code doesn't pass default gometalinter, we can just use a subset of linters if you wish, not sure what your plans on linting are

Lastly there is no Gopkg.toml so I had to manually go get packages. I'm all for not vendoring in libs, but shouldn't we have the Gopkg.toml here to declare what this depends on?


spenczar:
Thanks for doing this! Fixes #5 when it lands.

> First the readme says go 1.7+ but some of the helpers use t.Helper() which is a go1.9 feature. We can use compile pragmas to stop those tests from testing in 1.7 or we can change travis to only run in 1.9+

I think we still want to run tests in 1.7. We should fix the tests to be compatible across versions. Let's do that in a separate PR. I've made #27 for this.

> Second I didn't add linting because the code doesn't pass default gometalinter, we can just use a subset of linters if you wish, not sure what your plans on linting are.

It would probably be good to pass gometalinter when we can, but I agree with turning it off for now.

> Lastly there is no Gopkg.toml so I had to manually go get packages. I'm all for not vendoring in libs, but shouldn't we have the Gopkg.toml here to declare what this depends on?

Yes, 100%. We should include just a Gopkg.toml. Again, would be best in a separate PR.

spenczar:
#29 and #30 are merged, so this is ready for another pass. Please change the `go get` calls to `dep ensure -vendor-only`, and let's make sure the tests pass in Travis.

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+- tip
+install:
+- go get github.com/gogo/protobuf/...
+- go get github.com/golang/protobuf/...
+- go get github.com/pkg/errors/...
+- go get github.com/stretchr/testify/...
+- go get golang.org/x/net/...
+script:
+- go install ./...
+- go test -race ./...
```
spenczar: I'd like this to be `make test_all` so we run tests of client compatibility. But getting there would probably take a bunch more work - we'd need to install Python and protoc, etc. I would rather merge this, get CI going, and then improve it in a second pass, so let's delay that.

.travis.yml
```diff
@@ -0,0 +1,13 @@
+sudo: false
+language: go
+go:
+- 1.7
+- 1.8
+- 1.9
+- tip
+install:
+- go get github.com/golang/dep
```
dvic: I believe this should be `go get github.com/golang/dep/cmd/dep`?

spenczar:
Gareth and I discussed this over slack. We decided to just get this going with go1.9 and tip for now. We'll work our way back to go1.7+, which I want to support, but it's more important to just get Travis up and running for now.

GarethLewin:
We have a ✅ , pr is ready

.travis.yml
GarethLewin: Indeed, but we decided to just use the binary as per https://github.com/golang/dep/blob/master/docs/FAQ.md#how-do-i-use-dep-in-ci

PR #30, opened 2018-01-18
Record test dependencies with dep

### LINKED ISSUE #17: 'make' depends on state of GOPATH
[CONTRIBUTING.md](https://github.com/twitchtv/twirp/blob/master/CONTRIBUTING.md) states:
```Generally you want to make changes and run make, which will install all dependencies we know about...```
but I experienced build failures (with no changes).  However, running `dep init` to vendor dependencies first led to a successful run of `make`, suggesting something in my `GOPATH` was incompatible/outdated and not updated by `make`.

- Are there additional dependencies to install/update via `make`?
- Alternatively, could a `dep` manifest and/or vendor directory be used to stabilize dependencies? (or would that clash with the rest of `make`, which is built around `GOPATH` and `go install`?)

spenczar:
The main twirp library does not have any dependencies, but its tests and generated code examples do.

In order to run tests in a CI environment, we need a manifest listing our dependencies. This adds that list, but does not commit the vendor directory.

To avoid confusing the dep tool with false constraints, use overrides. There's a comment in Gopkg.toml more fully explaining this (thanks for the idea, @carolynvs!). 

Fixes #17, related to #25. 


GarethLewin:
Note that once this is in we can change #25 to just call `dep ensure -vendor-only`

Issue #19, opened 2018-01-17
advice on other langs


rileytg:
i’m interested in ruby and java support (maybe php too). any advice along these lines?

gaffneyc:
I've started on a generator for Ruby: https://github.com/gaffneyc/protoc-gen-twirp_ruby

It's functional enough to generate a working client based on the haberdasher demo. I would appreciate any feedback or pull requests. I'm prototyping a service with Twirp and will need it to talk the service from our main Rails app on Dead Man's Snitch. If the prototype is successful I'll keep working on it and productionatize it as I go.

spenczar:
Glad to hear it (and amazed at how quickly you got that out, @gaffneyc, holy smokes)!

I'd love to help. I think generating a client should be simple. Here's what I suggest:
1. Write a protobuf spec for a service. Make sure it has multiple methods, a few different message types - you want to cover some corner cases.
2. Generate Twirp server code and run a Twirp server on localhost so you have something to test against.
3. Write a client in the language you're targeting by hand. Try to make it feel nice to use, good and idiomatic.
4. Copy `protoc-gen-twirp`, give it a language-specific name like `protoc-gen-twirp_java` or whatever, and hack at your copy until it generates an output that looks just like your hand-written client.
5. Make sure your generated client works by testing it against the running Go server. You need to handle routing and serialization, and also error responses.
6. To make sure you handle errors correctly, run your client against `clientcompat`, which is a package in the Twirp repo that has a kind of framework for stressing a client implementation.

Your generated client should speak protobuf. It doesn't need to be able to speak JSON, really.

Some of the danger spots to watch out for:
 - Do you deserialize errors correctly?
 - Do you handle things correctly if there is no twirp server running at the address requested, and you just get (say) a generic nginx text error?
 - Does your generator correctly handle message definitions that are imported from other protobuf files?

As always, feel free to ping me directly if you need any help. The gophers slack is a good place to do this, or here.

gedw99:
Dart team tend to use JSON RPC.
https://github.com/dart-lang/json_rpc_2

gaffneyc:
@spenczar Thanks! The python generator was a really good reference that made it a lot easier. I had to reinvent a few wheels since the `gen` package is internal. In this case it was a good thing since it gave me the excuse to learn Go's templating.

daroot:
Would there be any interest in PRs for server-side support to the non-go generators?

[...]

PR #31, opened 2018-01-19
Evening Hack: generate browser-compatible javascript

Xe:
This generates reasonably sane javascript that is fairly universally compatible. Usage is simple:

Given this protobuf:

```protobuf
// printerfacts.proto
syntax = "proto3";

package us.xeserv.api.printerfacts;
option go_package = "proto";

service Printerfacts {
      rpc Fact(FactParams) returns (Facts) {};
}

message FactParams {
      int32 count = 1;
}

message Facts {
      repeated string facts = 1;
}
```

this tool will generate output javascript such as
https://xena.greedo.xeserv.us/files/twirp_printerfacts/printerfacts_twirp.js

usage is simple:

```shell
#!/bin/sh
# regen.sh
# regenerates protocol buffer generated code

protoc  --proto_path=$GOPATH/src:. \
        --twirp_out=. \
        --twirp_jsbrowser_out=. \
        --go_out=. \
        ./printerfacts.proto
```

Usage from HTML is simple:

```html
<script src="https://xena.greedo.xeserv.us/files/twirp_printerfacts/printerfacts_twirp.js"></script>
<script>
 let onSuccess = (x) => {
     document.write(x.facts[0]);
 };
 let onError = (x) => {
     console.log(x);
 };

 Printerfacts_fact("https://printerfacts.herokuapp.com", {"count": 1}, onSuccess, onError);
</script>
```

Ref #19

See example [here](https://printerfacts.herokuapp.com). (source code at https://github.com/Xe/printerfacts)


amerine:
Nice

spenczar:
This is very, very cool, and impressively lightweight. Thanks for sharing it!

I think I'd like to keep the bar for inclusion in this repo pretty high. Clients here should do all of the following:
 - feel very idiomatic to experts familiar with the language they're targeting
 - minimize external dependencies
 - speak protobuf to the server, not just JSON
 - play well with the existing official protobuf compiler for the language, if one exists
 - handle errors correctly

I can't say whether this is idiomatic - I'm not really a JavaScript pro. It clearly doesn't have any dependencies. But I'd like to see it speak in protobuf before accepting it.

There's an [official JavaScript protobuf compiler](https://github.com/google/protobuf/tree/master/js) these days. I think a generated JavaScript client should depend upon its output. 

It looks like the official compiler has options that change the output, letting you pick how you want to import the generated code. The official Twirp JS generator should accept identically-named options to be compatible.

I want to reiterate that this is just the bar for getting included in github.com/twitchtv/twirp. I don't think you should feel the bar is so high to host your generator elsewhere, but I'd like to keep the ones in here polished (and in fact the Python one needs a lot more polish right now).

In order to make it easier for others to write generators and host them elsewhere, I'm strongly considering making `github.com/twitchtv/twirp/internal/gen` public, moving it to `github.com/twitchtv/twirp/protocgen` or something.

joneskoo:
I'm also looking to use twirp with a client from browser without dependencies - for a simple API I'm fine with a JSON only client.

I would also like the typescript definition file to be generated. That way it's possible to have amazing quality of life as the need to write JavaScript to include a web frontend for a Go project is minimal, and my editor can tell me what fields are available and their types.

If this doesn't fit the twirp proper, the twirp documentation/README should have a blurp link to how to find these extensions.

Xe:
@joneskoo I've moved this binary into https://github.com/Xe/twirp-codegens, I don't personally like typescript but I'd be more than happy to accept a PR that generates typescript type definitions.

thechriswalker:
@spenczar I took your suggestions and produced this: https://github.com/thechriswalker/twirp/commit/c379f7e41fee1261387b3489b8d8292802f311ec

I haven't made a PR yet, but would be interested in feedback. It takes the official google protobuf javascript compiler output and creates the twirp api client on the side. It generates code for both node and the browser, speaking JSON and protobuf and the node version passes the `clientcompat` tests.

It does only generate commonjs modules currently.

spenczar:
https://github.com/twitchtv/protogen now exists, so you can import Twirp's protoc plugin library. Hooray!

I'm going to close this issue as community-maintained generators seem like the best way forward right now.

PR #34, opened 2018-01-19
Adding a third party license file

hyandell:
*Description of changes:*

Adding a third party license file to show the other licensing inside the project


THIRD_PARTY
```diff
@@ -0,0 +1,25 @@
+** Protobuf -- http://github.com/gogo/protobuf
```
spenczar: I think this should be https://github.com/golang/protobuf and its associated license.

We invoke `github.com/gogo/protobuf` for a test, but we don't distribute it, and nothing in Twirp is based on its code.

hyandell: Apologies - I picked that up from a repository search for protobuf and didn't realize there was a fork in the _tools directory :) Will fix.

PR #38, opened 2018-01-21
Allow Hook to access de-serialized request object

cperfect:
Adds a new hook method RequestDeserialized that allows the hook to access the deserialized request object before the service method is called. 

Uses for this: detailed logging/metrics, fine grained permissions, env overrides/additions to request

Includes associated code gen, tests and example changes.

Alters gitignore to allow use of dep and vendoring locally

*Issue #, if available:*

*Description of changes:*


spenczar:
I don't understand when this would be necessary, and I think it increases complexity for any readers - especially if the hook permits users to mutate the request. That makes it much harder to read the code linearly, because a hook might muck with stuff in some complex way. This is kinda like Rails' `:before_filter`, which are the kind of complex nightmare I want to avoid as strenuously as possible.

Further, I think that all the uses you mentioned ought to be done in the application, not through hooks. At the point when `RequestDeserialized` is called, all that happens next is the actual invocation of the application method. "Fine grained permissions" that depend on the actual request payload should be done within the application's actual method calls, not through hooks. "Logging" can be done by setting context fields in your application's method calls and pulling them out in the `ResponseSent` hook, which is probably when you want to be logging anyway. "Env overrides" similarly ought to be method-specific.

One sign that hooks don't work as a design here is that you're forced to use `interface{}`. That would mean that any code that used `RequestDeserialized` would need to assert the request payload object to the right type to be able to do anything - it would need to do `reqobj.(MakeHatRequest)`, or something. But how will it know the right type to pick? It will need to decide based on the method, or just try many different types in a big switch. In either case, it will have to be logic specific to the type... but you already will have that information in the actual application.

Thank you for being thorough and writing tests and fixing examples, but I don't think this adds enough to be justified. Feel free to open an issue if you'd like to discuss the idea further.

PR #39, opened 2018-01-21
Replace *http.Client in client constructors with new HTTPClient interface

### LINKED ISSUE #12: Introduce HTTPClient interface
It would be great to remove dependency to `http.Client` type. Instead of this interface can be used:
```go
// HTTPClient sends http.Requests and returns http.Responses or errors in case of failure.
type HTTPClient interface {
	Do(req *http.Request) (*http.Response, error)
}
```

It allows creating decorators for standard http.Client with some additional functionality like metrics, backoff retries, etc. `http.Client` implements this interface and such code is backward compatible with the current implementation. We use such approach and it's great.

What do you think? If you like this idea I can make PR.

nawuko:
*Issue #, if available:* #12 

*Description of changes:* Proof-of-concept work for resolving issue #12 

*ToDo:*
- [x] Check implementation since this is PR is from a GoLang beginner.
- [x] Replace the `type HTTPClient interface` comment with proper english.
- [x] Check if the name for the interface `HTTPClient` is good or should it be changed
- [x] Is the placement for the `HTTPClient` interface acceptable. Currently in the `generateUtils` block

I have tested this PR with following clients:
- http.Client
- https://github.com/sethgrid/pester (wrapper of http.Client)


protoc-gen-twirp/generator.go
```diff
@@ -294,6 +294,13 @@ func (t *twirp) generateUtilImports() {
 // These should be generated just once per package.
 func (t *twirp) generateUtils() {
 	t.sectionComment(`Utils`)
+	t.P(`// HTTPClient is the interface used for HTTP Requests. If you use something else`)
+	t.P(`// than the default http.Client you lose twirps redirect policy. See CheckRedirect implementation`)
+	t.P(`// We highly recommend using the same implementation for correct error handling and compatibility`)
```
spenczar: I'm a little nitpicky on documentation. Could you revise to the following:
```go
t.P(`// HTTPClient is the interface used by generated clients to send HTTP requests.`)
t.P(`// It is fulfilled by *(net/http).Client, which is sufficient for most users.`)
t.P(`// Users can provide their own implementation for special retry policies.`)
t.P(`// `)
t.P(`// HTTPClient implementations should not follow redirects. Redirects are`)
t.P(`// automatically disabled if *(net/http).Client is passed to client`)
t.P(`// constructors. See the withoutRedirects function in this file for more`)
t.P(`// details.`)
```

nawuko:
@spenczar I regenerated all proto *.twirp.go files otherwise i couldnt test against them. Implemented a custom test not 100% sure about the name. Updated the comment aswell.

internal/twirptest/client_test.go
```diff
@@ -299,3 +326,12 @@ func errCause(err error) error {
 	}
 	return cause
 }
+
+// wrappedHTTPClient is a http.Client which is wrapped to test against the HTTPClient interface
```
nawuko: This really should be changed to good english ...

internal/twirptest/client_test.go
```diff
@@ -282,6 +282,33 @@ func TestClientErrorsCanBeCaused(t *testing.T) {
...
+func TestCustomHTTPClientInterface(t *testing.T) {
+	// Start up a server just so we can make a working client later.
+	h := PickyHatmaker(1)
+	s := httptest.NewServer(NewHaberdasherServer(h, nil))
+	defer s.Close()
+
+	// Create a custom wrapper to wrap our default client
+	httpClient := &wrappedHTTPClient{
+		Client: http.DefaultClient,
+	}
```
spenczar: This is a bit pedantic, but I want to be careful. We should have an assertion somewhere in here that we call the right `httpClient`.

Could you change `wrappedHTTPClient` to support a `wasCalled() bool` or something similar, and then check that we did call it? This test will obviously pass today, your code is quite clear, but I want to be protected from breaking it accidentally if we add more logic into the client constructor in the future.

internal/twirptest/client_test.go
spenczar: Eh, it's not too bad. It's just an unexported type in test code. Maybe something like:
```go
// wrappedHTTPClient implements HTTPClient, but can be inspected during tests.
```

internal/twirptest/client_test.go
```diff
@@ -299,3 +326,12 @@ func errCause(err error) error {
 	}
 	return cause
 }
+
+// wrappedHTTPClient is a http.Client which is wrapped to test against the HTTPClient interface
+type wrappedHTTPClient struct {
+	Client *http.Client
```
spenczar: No need for this to be exported - change to `client *http.Client`.

PR #40, opened 2018-01-21
Move package gen to experimental package set

Xe:
This PR closes #31.

**Description of changes**

Package `gen` is incredibly useful for writing twirp client generators. This patch exposes package `gen` in the experimental prefix `/exp/`. Think of `/exp/` as halfway between internal and public code. No stability guarantees apply.


Fugiman:
While I think there's value in making `gen` public, and agree that marking it as "experimental" for a period of time is a good idea, I worry that encoding that into the path forces us to make a backwards incompatible change in the future to make it public. I wonder if it'd be better to not put it in an `/exp/` package, but instead mark it as not being stable via documentation. It may also be worth making an entirely new repo for it if we want to support semantically versioning it seperate from Twirp.

I leave it to @spenczar to make the final verdict.

spenczar:
I agree with @fugiman that putting this in a different repository seems to be clearly better. I'll have to just double-check with our lawyers that they're cool with that, but let's go that route. I like it more because:

 - We can highlight the instability of the package loudly in the README.
 - We can version the package separately.
 - I can triage PRs and bug reports separately.

The only downside is that `protoc-gen-twirp` will need to vendor it, and development across two repos can be a little clunkier, but that's not so bad.

I hope to have an answer back from the lawyers in the next two days or so. Thanks for bearing with me, @Xe :)

spenczar:
We've released https://github.com/twitchtv/protogen to help meet this need. Hooray!

Issue #18, opened 2018-01-17
Makefile does not enforce a particular protoc version, so generated files churn


jmank88:
Running `make` from master/v5.0.0 (db96cdf354e8dc053e5ee5fe890bb0a7f18123ab) without any changes modifies 9 python test files.  I expected no changes to checked-in files.

Each includes additional code to register a service descriptor, similar to:
```python
_SVC = _descriptor.ServiceDescriptor(
  name='Svc',
  full_name='twirp.internal.twirptest.proto.Svc',
  file=DESCRIPTOR,
  index=0,
  options=None,
  serialized_start=54,
  serialized_end=141,
  methods=[
  _descriptor.MethodDescriptor(
    name='Send',
    full_name='twirp.internal.twirptest.proto.Svc.Send',
    index=0,
    containing_service=None,
    input_type=_MSG,
    output_type=_MSG,
    options=None,
  ),
])
_sym_db.RegisterServiceDescriptor(_SVC)

DESCRIPTOR.services_by_name['Svc'] = _SVC
```

Should these files be updated with these changes?

spenczar:
Hm, this probably indicates that you have a different version of `protoc` than I do - I don't see any changes when I run `make`. 

This is unfortunate. It's a little hard to pin the version of `protoc` and keep it the same across invocations. We might need to revisit the `install_proto.sh` script that the Makefile invokes and lock things down tighter.

jmank88:
It looks like I'm running `3.4.0` but the script would otherwise install `3.1.0`.

jmank88:
The generated Go files have begun to drift out of sync as well - #52 as a reference.  It would be neat for CI to run `make` (or at least generation) and disallow modifications.

spenczar:
I'm not sure I'd call this _fixed_ yet; we're going to continue to drift in the future. I think it will be fixed when `make` enforces a particular `protoc` version.

I'm leaning towards saying `make` should just check `protoc --version` and yell at you if it doesn't exactly match a blessed version (`3.5.1` seems good, I guess?).

spenczar:
Fixed in #63.

PR #43, opened 2018-01-22
remove done func in clients

mkabischev:
*Description of changes:*

According to context's [documentation](https://golang.org/pkg/context/#Context):
```
// If Done is not yet closed, Err returns nil.
// If Done is closed, Err returns a non-nil error explaining why:
// Canceled if the context was canceled
// or DeadlineExceeded if the context's deadline passed.
// After Err returns a non-nil error, successive calls to Err return the same error.
Err() error
```

So it means that `done` function can be replaced with simple `err := ctx.Err(); err != nil`.


mkabischev:
Some *.py files have been changed during running tests. Can you suggest what version of protoc do you use?

spenczar:
@mkabischev We're kind of all over the place on the `protoc` version - that's #18. I think the current stuff was generated with 3.3.0.

If there's a good technique for pinning `protoc`'s version, I'd love to hear about it on #18. For now, I'll just accept the noise from the churn of that file.

Issue #32, opened 2018-01-19
Twirp on AWS APIG+Lambda


rynop:
Anyone tried to run Twirp on Lambda yet?

I'm not a fan of API Gateway (APIG) and all the complexity, but I love lambda.  I think it would be sweet to do the following:

1. Configure APIG to handle `ANY` method on the `/{proxy+}` resource will reverse proxy everything to **one** Lambda function (lambda function is a go binary using Twirp).
1. HTTP 1.1 `application/json` request comes into APIG, it gets routed to Lambda
1. Thin go handler code transforms the `events. APIGatewayProxyRequest` ([github.com/aws/aws-lambda-go/events](https://github.com/aws/aws-lambda-go/tree/master/events)) to the HTTP request object that Twitch needs.
1. Twitch router works as normal
1. Thin layer that takes the Twirp response and transforms it to a `return events.APIGatewayProxyResponse`

Questions (I'm a go noob, never used RPC or protobuf but they make sense to me):

1. Am I missing anything big here that would prevent me from doing this?
1. Can anyone point me to the Twirp code that would get me started on 3 and 5 above?

rynop:
https://github.com/apex/gateway looks to be promising, will investigate and report back

rynop:
https://github.com/apex/gateway still requires a full blown HTTP server (listening on a port etc).  I think I can get it to work, but my “ideal” solution would be to use all the Twirp badassness (routing, code generation, error handling etc) WITHOUT the need for the HTTP stack because APIG is already providing the HTTP stack - for example already parsed the headers out of HTTP, which apex/gateway [has to re-do](https://github.com/apex/gateway/blob/master/request.go#L49) to send onto `net/http`.  

Would it require a ton of Twirp re-factoring to “turn off” the http protocol stuff and instead just pass the Headers and POST body to Twirp directly?  I’d just pass along the Twirp sterilized JSON response on the Lambda `return`. 

Bummer is to get `application/protobuf` working I’d have to base64 encode the Twrip protobuf response before sending it back from Lambda (which incurs performance penalty)

rynop:
@spenczar correctly pointed out to me that: 

> the `address` part of apex/gateway's ListenAndServe is a red herring - if you check the source, you'll see that it's unused. It's just there to match the standard library `net/http.ListenAndServe` signature that people are used to

[...]

PR #46, opened 2018-01-22
Add protobuf support for AWS API Gateway

rynop:
In making [Twirp work on APIG+Lambda](https://github.com/twitchtv/twirp/issues/32) I discovered that protobuf was not working even when I added `application/protobuf` to APIG settings for binary.

Long story short, for binary support to work in APIG via Lambda Proxy, the client (in this case Twirp go Protobuf client) MUST send the `Accept: application/protobuf` header.

Full AWS docs on this topic can be found [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format).

I can't think of any negative implication for always adding `Accept: application/protobuf` request header when using a `New<Service>ProtobufClient`.  Since errors always return `application/json` content-type (nice design BTW), the error path is handled cleanly on APIG if you specify `application/protobuf` as a binary response type.

Sorry I tried to run your tests but got some errors:
```
~/g/s/g/r/twirp (master|✔) $ make                                                                                                                                                                   15:16:35
./install_proto.bash
/usr/local/bin/protoc
Protoc already installed
GOPATH=/Users/ryan/go/src/github.com/rynop/twirp/_tools go install github.com/twitchtv/retool/...
/Users/ryan/go/src/github.com/rynop/twirp/_tools/bin/retool build
retool: installing github.com/golang/protobuf/protoc-gen-go
retool: installing github.com/kisielk/errcheck
retool: installing github.com/gogo/protobuf/protoc-gen-gofast
retool: installing github.com/twitchtv/retool
PATH=<REMOVED>
GOBIN="/Users/ryan/go/src/github.com/rynop/twirp/bin" go install -v ./protoc-gen-...
protoc-gen-twirp/generator.go:32:2: use of internal package not allowed
protoc-gen-twirp/generator.go:33:2: use of internal package not allowed
protoc-gen-twirp/generator.go:34:2: use of internal package not allowed
protoc-gen-twirp_python/main.go:28:2: use of internal package not allowed
protoc-gen-twirp_python/main.go:29:2: use of internal package not allowed
protoc-gen-twirp_python/main.go:30:2: use of internal package not allowed
make: *** [generate] Error 1
~/g/s/g/r/twirp (master|✔) $ go version                                                                                                                                                             15:16:42
go version go1.9.3 darwin/amd64
```


spenczar:
First the tests: You are unable to run tests because you have the project checked out in `$GOPATH/src/github.com/rynop/twirp`, but all the code inside uses imports that point to `$GOPATH/src/github.com/twitch/twirp`. This includes an attempt to import of `github.com/twitch/twirp/internal/gen`, which is an internal package, which the go tool is forbidding you from doing.

The right thing here is to do your work in `$GOPATH/src/github.com/twitch/twirp`. To work with a fork, use `git remote add fork git@github.com:rynop/twirp.git`; then you should be able to push your changes to your fork (with `git push fork <branch>`, or whatever) but still have the code checked out in the right part of GOPATH.

---

Second, your suggestion about `Accept`: I can't seem to find the precise place where this is documented in API Gateway. I'm having some trouble wading through the docs, but it sounds alittle like you can configure the gateway to pass through a binary `application/protobuf` response if it's configured correctly? I'm basing that off of [this page](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html).

I'm hesitant to add `Accept` because I wouldn't want it to be a requirement in the spec for Twirp clients, and I can see us slipping into requiring it accidentally if we put it in the Go client. I could be convinced, but I want some clearer evidence that this is required by infrastructure like API Gateway.

For now, you could do this by hacking up your `http.RoundTripper` used in the `*http.Client`. When #39 lands, you'd be able to do it through a wrapper there. Kind of unpleasant, though.

rynop:
Yea sorry I included the wrong link.  Docs are not super clear, check out the [API Gateway Response Content Type Conversions](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html) for a little more detail.  If you use Lambda proxy integrations "you do not have any option to modify the passthrough behaviors" ([source](https://docs.aws.amazon.com/apigateway/latest/developerguide/integration-passthrough-behaviors.html) - therefore the `contentHandling` is `Undefined`.

Therefore a "a binary type" Request `Accept` header is required to get a binary payload response.

I did some research on the topic back in October, detailed findings [here](https://rynop.com/2017/10/17/aws-api-gateway-binary-support-using-lambda-real-world-example-image-resize/) if you care.

I'm guessing/hoping Twitch in a serverless environment becomes popular, especially if we make it easy.  Since AWS is market leader, IMO it would be nice to have the `Accept` header auto added in the protobuf request path.  I can see your hesitation on not wanting it to creep into the Twirp spec as a requirement.  I'm very anti extra config for common use cases, but would it be possible to add some sort of `--for-api-gateway` flag to `protoc` -or- some option to `service.proto`?  When run it would do something along the lines of what is included in this PR?

Thanks for the tip on local dev.  Just started learning go last week, so hopefully I don't lean on you too much here during my learning curve.  Based on your response to my question I'll spend time on getting the tests to pass.

rynop:
Was looking into how to hack up `http.RoundTripper`.  [Their docs](https://golang.org/pkg/net/http/#RoundTripper) explicitly say not NOT do what you are proposing:

>// RoundTrip should not modify the request, except for
>// consuming and closing the Request's Body.

However, I ended up finding another way per your WIKI on HTTP headers.  If I attach the `Accept` header to the ctx that my HTTP client uses, it works:

```golang
        ...
        header := make(http.Header)
	header.Set("Accept", "application/protobuf")
	ctx := context.Background()
	ctx, err := twirp.WithHTTPRequestHeaders(ctx, header)
	if err != nil {
		log.Printf("twirp error setting headers: %s", err)
		return
	}

       resp, err := client.CreateGiphy(ctx, &image.Search{Term: "wahooo"})
       ...
```

I'm still not a huge fan of doing this manually for every HTTP client, but it is definitely workable.

I won't be offended at all if you close this PR and give the guidance of going with this latest approach.  Thanks.

spenczar:
Right, it's kind of blechy right now. I hope with #39 it will be easy to provide a library that has a valid HTTPClient for talking to API Gateway.

> Was looking into how to hack up http.RoundTripper. Their docs explicitly say not NOT do what you are proposing

It's forbidden to alter the memory of an `http.Request`, but you can copy it and make changes on your copy. Easiest way to do that is with `req.WithContext`, which makes a valid copy.

Your solution using `WithHTTPRequestHeaders` works too, though.

Regarding the original proposal: yeah, I'd like to hold off on adding a new header to all our outbound requests until we have a _really_ good reason to do so. I imagine that such a change would require a version bump. Let's close this for now, but we'll keep it in mind if a similar request appears in the future.

Issue #15, opened 2018-01-17
Wrong import generation for external package


Ganitzsh:
Hi there!

Thanks for bringing this framework to us 👍 

I've been playing around with it for a bit today, and it looks like the generation of the imports has an issue. Either that or I'm just missing something so I hope you can help me out!

I'm generating the `.twirp.go` file using this command line:

`protoc -I . --twirp_out=./out --go_out=./out service_def.proto dep1.proto dep2.proto...`

In the generated `twirp` service file I have the following import:

`import mypackagename "."`

But this cannot work when I want to use the generated source from an external package, and of course I get the following error: `local import "." in non-local package` on my external package (implemetation) when I import a message type in my source code.

The upper part of my service is as follows:
```
syntax = "proto3";

package some.thing;
option go_package = "mypackagename";
```

spenczar:
This looks like it could certainly be a bug in our handling of dot-imports within a package, they can be tricky.

Can you post a reproducing case with the complete text of the proto files?

polds:
Sorry for hijacking but I just ran into I think the same issue while trying to import "google/protobuf/descriptor.proto";

Reproducible:
```Protobuf
syntax = "proto3";
import "google/protobuf/descriptor.proto";
package twitch.twirp.example;
option go_package = "example";

extend google.protobuf.MessageOptions {
    string some_key = 1000;
}

message Hat {
  option (some_key) = "a hat!"
  int32 size = 1;
  string color = 2;
  string name = 3;
}

message Size {
  int32 inches = 1;
}

service Haberdasher {
  rpc MakeHat(Size) returns (Hat);
}
```

Yields the following Go import:
```Golang
import proto "github.com/golang/protobuf/proto"
import fmt "fmt"
import math "math"
import google_protobuf "google/protobuf"
```

Which obviously yields the error:
`cannot find package "google/protobuf"`

I verified that importing something like `google/protobuf/timestamp.proto` **does** work correctly, it's just the descriptor.proto that doesn't.

spenczar:
@polds Can you make a separate issue? I am almost completely certain that that has a different root cause. The Go protobuf compiler has some [fiddly behavior](https://github.com/golang/protobuf/issues/143) around the "[Well Known Types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf)", which I think you are running into.

Ganitzsh:
@spenczar Okay so apparently this dot import is showing only when importing local definitions.

I tried the following:

[...]

PR #47, opened 2018-01-22
protoc-gen-twirp: skip Go import for proto files in the same package

rhysh:
The Go interface generator makes sure that all messages that are used in RPCs are imported, but it can go too far by attempting to write a Go `import` line for protobuf messages that are defined in a different protobuf file in the same protobuf package.

Instead of checking if the protobuf filenames match to determine whether a Go `import` line is required, check if the protobuf packages (which line up with Go packages) match.

Fixes https://github.com/twitchtv/twirp/issues/15


RoarkeRandall:
I'm still getting this error: 
```
service.twirp.go:26:8: local import "." in non-local package
```

after running 

```
go get github.com/twitchtv/twirp/protoc-gen-twirp
```

I have a messages.proto, and a service.proto that imports "messages.proto" (they are in the same directory)

RoarkeRandall:
Sorry, I needed a -u on my go get. issue is fixed.

PR #49, opened 2018-01-23
adding context-metadata to client

### LINKED ISSUE #45: Add context metadata to generated client's request context
Hi, can you suggest what load balancing/service discovery strategies do you use in twitch? 
As I see the current implementation of twirp allows only to use server-side discovery using some reverse-proxy like ELB, nginx, hproxy, etc. But it will be great to have the possibility to use client-side discovery as well. 
It becomes easier to implement it with #39 because I can replace URL with the discovered host in the request. So imagine I can write some custom implementation of `HTTPClient` interface or `http.RoundTripper` like:

```go
// ClientWithDiscovery is an implementation of HTTPClient interface that automatically discovers services' hosts using consul/etcd/...
type ClientWithDiscovery struct {
   service string
   client    HTTPClient
}

func (c *ClientWithDiscovery) Do(r *http.Request) (*http.Response, error) {
    r.URL.Host = c.pickHostForService()
    return c.client.Do(r)       
}
```

The big problem of such implementation is that service is a property of the client and I have to create separate client for each service I want to use. But it can be easily solved if in request's context there will be information about service. Client generated code can contain lines like:
```go
func (c *haberdasherJSONClient) MakeHat(ctx context.Context, in *Size) (*Hat, error) {
        ctx = ctxsetters.WithServiceName(ctx, "Haberdasher") // <-- new line
	url := c.urlBase + HaberdasherPathPrefix + "MakeHat"
	out := new(Hat)
	err := doJSONRequest(ctx, c.client, url, in, out)
	return out, err
}
```

So then I can call `twirp.ServiceName(ctx)` in my new client to fetch service name and make discovery. 

What do you think?

mkabischev:
*Issue #, if available:* #45 

*Description of changes:*

Adding some metadata to context passed to httpClient.


mkabischev:
@spenczar, sorry for delay. I've added test as you wanted.

PR #53, opened 2018-01-23
go-client: precompute service url strings

jmank88:
This PR proposes a small change to generated Go clients. By replacing the `urlBase` field with a pre-computed array of service `urls`, service methods can do an array look-up instead of repetitive string concatenation, at the cost of more initialization in the constructor functions.

### DIFF
--- a/clientcompat/internal/clientcompat/clientcompat.twirp.go
+++ b/clientcompat/internal/clientcompat/clientcompat.twirp.go
@@ -43,30 +43,32 @@ type CompatService interface {
 // =============================
 
 type compatServiceProtobufClient struct {
-	urlBase string
-	client  *http.Client
+	client *http.Client
+	urls   [2]string
 }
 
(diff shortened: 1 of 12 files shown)


clientcompat/internal/clientcompat/clientcompat.twirp.go
```diff
@@ -43,30 +43,32 @@ type CompatService interface {
...
+			prefix + "NoopMethod",
+		},
 	}
 }
 
 func (c *compatServiceProtobufClient) Method(ctx context.Context, in *Req) (*Resp, error) {
-	url := c.urlBase + CompatServicePathPrefix + "Method"
 	out := new(Resp)
-	err := doProtoRequest(ctx, c.client, url, in, out)
+	err := doProtoRequest(ctx, c.client, c.urls[0], in, out)
```
jmank88: If these literal numbers are too opaque, they could be swapped for `iota` `const`s, or the whole array could be swapped for a flat list of `methodNameURL string` fields.

Issue #61, opened 2018-01-26
List community-created generators on the README


spenczar:
There has been a really incredible burst of community-written generators. I want to highlight them and send people towards them.

Let's add a bit to the README, close to the top, that lists the generators out there for different languages.

I think the standard here should be that we list generators that make valid a valid client from a protobuf file, will be actively maintained, and which are in non-go environments. It can be a pretty wide list right now. **In the future, we may get more picky**, but for now, let's just list what's out there.

thechriswalker:
I created: https://github.com/thechriswalker/protoc-gen-twirp_js for javascript in browser and nodejs.

It takes on board the comments on https://github.com/twitchtv/twirp/pull/31 and fulfils most your criteria in this issue (except possibly the non-go one).

I write this as a comment rather than a PR for a few reasons:
 1. Given you haven't made the section of the README yet, I wasn't sure how you would format it.
 2. I copied (not-quite-vendored) the `internal/gen` package on which I originally based this - I wasn't sure whether you did plan to publicise them.
 3. Your requirement `which are in a non-go environment`, I guess is for the codegen itself, rather than the generated code? (the generated code will almost certainly not be go!). This generator is written in go, so doesn't fulfil that bit.

Given the above: should I submit the repo as something for this list, or the code in my fork of twirp for this repo?

spenczar:
@thechriswalker Right, I wasn't super clear. I'll make the initial list.

`internal/gen` will get its own repo very soon. I just need to get a thumbs-up from lawyers on naming the new repo properly.

Heh, I meant that the generated code should not be in go - I mostly mean "we don't want to direct people towards alternate Go generators; instead we want to make the official go generator really good." You're right that this was probably obvious :)

I'll include your repo in the initial list.

spenczar:
Done in #62.

PR #62, opened 2018-01-26
List third party generators on the README

spenczar:
As laid out in #61, I'd like to highlight these community-written generators.

I'd like an explicit okay from each person I'm linking to:
- [x] @elliots
- [x] @fajran
- [x] @gaffneyc
- [x] @thechriswalker
- [x] @Xe

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,22 @@ You will also need:
 ### Documentation
 Thorough documentation is [on the wiki](https://github.com/twitchtv/twirp/wiki).
 
+### Implementations in other languages
+
+This repo only has the official generators, which write out Go and Python code.
+For other languages, there are third-party generators available:
+
+|    Language    | Repository |
+|----------------|------------|
+| **JavaScript** | [github.com/thechriswalker/protoc-gen-twirp_js](https://github.com/thechriswalker/protoc-gen-twirp_js)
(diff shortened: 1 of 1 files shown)


thechriswalker:
fine with me.

gaffneyc:
Yeah, sounds great to me.

cactus:
@spenczar Would it make sense to add another column to the table to outline whether the linked repo generates clients, servers, or both?

spenczar:
@cactus Right now, all of these just make clients. If that's unclear, sure, we can add to the table.

elliots:
Great :) 

This morning I'll test out a few of the more popular language clients, and add some docs for importing the swagger files into AWS API Gateway, Azure API Management etc. 

P.S This one can make servers for ada, erlang, go, node, rush, scala... but i have no idea if they're any good. And jmeter config for benchmarking. And a million other mostly useless things. Still all json-only though.

fajran:
Could you also add mine? https://github.com/fajran/protoc-gen-twirp_java_jaxrs

At the moment it can already generate client for java using jaxrs library. I plan to create the server part as well 

Thanks

Xe:
Fine with me!

spenczar:
Rad, thanks everyone!

PR #64, opened 2018-01-26
Commit vendor

### LINKED ISSUE #59: Vendor directory is not committed, so protoc-gen-twirp does not build reliably.
https://github.com/twitchtv/twirp/blob/4492f81f75b4d504a2eadbda05055f416c076858/Gopkg.toml#L2

Mentions that these deps are only needed for testing. This is however, untrue. They are also needed to build the generators.

spenczar:
*Issue #, if available:*

Related to #17, #30, #57, #59. Fixes #59.

*Description of changes:*

Commit vendored code to the repo. This will let people `go get github.com/twitchtv/twirp/protoc-gen-twirp` safely.

I don't think there is any downside because `github.com/twitchtv/twirp`, which is the only thing users import, doesn't actually import any non-stdlib dependencies. I verified this with `go list`. 

Along the way, I changed travis to get rid of its dep references.


GarethLewin:
> I verified this with go list.

Is it worth the time to change the travis code to verify this so that this doesn't become incorrect in the future?

spenczar:
@GarethLewin even better, a `testing.T` test case so you don't have to make a PR to know. Good idea.

Issue #55, opened 2018-01-24
Proposed new routing format: /package/service/method


spenczar:
I think we could improve Twirp's routing scheme. It has two problems right now: the `/twirp/` prefix, and the `package.service` scheme (instead of `package/service`).

Some people have mentioned to me that they're hesitant to use Twirp because the routes are prefixed with /twirp, which is concerning to them for several reasons:

 - Trademarks: some very large organizations don't want to take any legal risks and are concerned that "twirp" could become trademarked.
 - Feels like advertising: This was mentioned in #48: putting "twirp" in all your routes feels like it's just supposed to pump Twirp's brand.
 - Homophonous with "twerp": In some Very Serious settings (like government websites), it's not okay that "twirp" sounds like "twerp", which means something like "insignificant pest."

We currently have the `/twirp` prefix to avoid colliding with gRPC's routes, which live at `/package.service/method`, and to reserve a space for future APIs (like a reflection server). 

Slashes-based routes (instead of dot-based) would be preferable because lots of existing technology expects slash-based routing. Load balancers and proxies use slashes for namespaces. If Twirp used a slash after the package name, users could leverage this much more easily. This is often possible with dots, just annoying, as you have to write a (possibly pretty complex) regex.

I think we can fix both of these in one go. This would be a breaking change at the protocol level, so it's something to take very seriously, but I think we can do it without much pain. It would just take a version bump to v6.

---

#### Proposed Implementation:

I propose we use a new routing scheme: `/<package>/<service>/<method>`.

This does not collide with gRPC's routes, which is good. I think it gives us flexibility to put future routes in by putting them under a Twirp package. We could provide a reflection service like this one, which lists available services:

```protobuf
syntax = "proto3";

package twirp.meta

import "github.com/golang/protobuf/protoc-gen-go/descriptor/descriptor.proto"

service Reflector {
  rpc ListServices(ListServiceRequest) returns (ListServiceResponse);
}

message ListServiceRequest {}
message ListServiceResponse {
  repeated protobuf.ServiceDescriptorProto services = 1;
}
```

It would live at `/twirp.meta/Reflector`, which seems good to me.

Regarding compatibility: generated clients would always use the new routing scheme. Generated servers would be able to use both, in case your users have old generated clients still around. We would accomplish this by exporting two path prefixes:

```go
const (
  HaberdasherPathPrefix = "/notreal.example/Haberdasher/"
[...]

PR #71, opened 2018-01-30
Create PROTOCOL.md

mocax:
Draft the wire protocol spec for Twirp based on latest discussions. It reflects the existing Twirp wire spec, plus best practice from proto3, gRPC, HTTP, and JSON. The spec should be usable for developers to build clients and servers using Twirp wire protocol without much dependency.

*Issue #, if available:*
#71  
*Description of changes:*
Write up the Twirp wire protocol.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1,16 +1,7 @@
 # Twirp Wire Protocol
 
-This document defines the Twirp wire protocol over HTTP.
-
-## Conventions
-
-The requirement level keywords "MUST", "MUST NOT", "REQUIRED",
-"SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY",
-and "OPTIONAL" used in this document are to be interpreted as
-described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
-
-The grammar rules used in this document are using [ABNF
-syntax](https://tools.ietf.org/html/rfc5234).
+This document defines the Twirp wire protocol over HTTP. The
+current protocol version is v5.
 
 ## Overview
 
@@ -28,28 +19,30 @@ the client can communicate with the server by making RPC calls.
 
 The Twirp wire protocol supports both binary and JSON encodings of
 proto messages, and works with any HTTP client and any HTTP version.
-However, certain capabilities may be limited by the actual HTTP
-library being used.
 
 ### URLs
 
-**URL ::= Base-URL "/" [ Package "." ] Interface "/" Method**
+In [ABNF syntax](https://tools.ietf.org/html/rfc5234), Twirp's URLs
+have the following format:
 
-The Twirp wire protocol uses HTTP URLs to directly specify the RPC
-endpoints on the server for sending the requests. such direct mapping
+**URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
+
+The Twirp wire protocol uses HTTP URLs to specify the RPC
+endpoints on the server for sending the requests. Such direct mapping
 makes the request routing simple and efficient. The Twirp URLs have
 the following components.
 
 * **Base-URL** is the virtual location of a Twirp API server, which is
-  typically published via API documentation or service discovery.  For
-  example, "https://example.com/apis".
+  typically published via API documentation or service discovery.
+  Currently, it should only contain URL `scheme` and `authority`. For
+  example, "https://example.com".
 
 * **Package** is the proto `package` name for an API, which is often
   considered as an API version. For example,
   `example.calendar.v1`. This component is omitted if the API
   definition doesn't have a package name.
 
-* **Interface** is the proto `service` name for an API. For example,
+* **Service** is the proto `service` name for an API. For example,
   `CalendarService`.
 
 * **Method** is the proto `rpc` name for an API method. For example,
@@ -57,40 +50,22 @@ the following components.
 
 ### Requests
 
-**Request ::= Request-Headers Request-Body**
-
 Twirp always uses HTTP POST method to send requests, because it
 closely matches the semantics of RPC methods.
 
 The **Request-Headers** are normal HTTP headers. The Twirp wire
 protocol uses the following headers.
 
-* **Authorization** header is often used to pass user credentials
-  from the client to the server, such as OAuth access token or
-  JWT token.
-
 * **Content-Type** header indicates the proto message encoding, which
-  should be one of "application/x-protobuf", "application/json". The
+  should be one of "application/protobuf", "application/json". The
   server uses this value to decide how to parse the request body,
   and encode the response body.
-
-* **User-Agent** header indicates the client application and its
-  runtime environment. While the server should not use this
-  information for request processing, this header is heavily used
-  for analytics and troubleshooting purposes.
-
-* **RPC-Timeout** header indicates the client-specified request
-  timeout in seconds, such as "10". If **RPC-Timeout** is omitted, the
-  server should use a pre-configured timeout value, by default it
-  should be 10 seconds.
 
 The **Request-Body** is the encoded request message, contained in the
 HTTP request body. The encoding is specified by the `Content-Type`
 header.
 
 ### Responses
-
-**Response ::= Response-Headers Response-Body**
 
 The **Response-Headers** are just normal HTTP response headers. The
 Twirp wire protocol uses the following headers.
@@ -111,6 +86,8 @@ corresponding wire payloads.
 The example assumes the server base URL is "https://example.com".
 
 ```proto
+syntax = "proto3";
+
 package twirp;
 
 service Echo {
@@ -131,7 +108,7 @@ message HelloResponse {
 ```
 POST /twirp.Echo/Hello HTTP/1.1
 Host: example.com
-Content-Type: application/x-protobuf
+Content-Type: application/protobuf
 Content-Length: 15
 
 <encoded HelloRequest>
@@ -152,7 +129,7 @@ Content-Length: 27
 
 ```
 HTTP/1.1 200 OK
-Content-Type: application/x-protobuf
+Content-Type: application/protobuf
 Content-Length: 15
 
 <encoded HelloResponse>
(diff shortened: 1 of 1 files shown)


daroot:
`application/x-protobuf` or `application/protobuf`?   The code uses the latter, not the former.

slaskis:
Charles Proxy [seems to support](https://www.charlesproxy.com/documentation/using-charles/protocol-buffers/) either `application/x-protobuf` or `application/x-google-protobuf` and it's nice to have supported tooling. Although maybe using `application/protobuf` in Twirp might coax the Charles devs to add that Content-Type too...

mocax:
Google didn't complete its IETF application, https://tools.ietf.org/html/draft-rfernando-protocol-buffers-00. So technically it still is `application/x-protobuf`.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+To use Twirp, developers first define their APIs using proto files,
+then use Twirp tools to generate the client and the server libraries.
+The generated libraries implement the Twirp wire protocol, using the
+standard HTTP library provided by the programming language runtime or
+the operating system. Once the client and the server are implemented,
+the client can communicate with the server by making RPC calls.
+
+The Twirp wire protocol supports both binary and JSON encodings of
+proto messages, and works with any HTTP client and any HTTP version.
+However, certain capabilities may be limited by the actual HTTP
```
spenczar: I don't think it's worth including this caveat until we can actually say this precisely.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+library being used.
+
+### URLs
+
+**URL ::= Base-URL "/" [ Package "." ] Interface "/" Method**
```
spenczar: Please change `Interface` to `Service`.

Also, the current routing is `URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method`, please update this to match the `/twirp` prefix.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+The Twirp wire protocol uses HTTP URLs to directly specify the RPC
```
spenczar: remove "directly"

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+endpoints on the server for sending the requests. such direct mapping
```
spenczar: Capitalize "direct"

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+makes the request routing simple and efficient. The Twirp URLs have
+the following components.
+
+* **Base-URL** is the virtual location of a Twirp API server, which is
+  typically published via API documentation or service discovery.  For
+  example, "https://example.com/apis".
```
spenczar: This is a bad example right now, as the base URL in v5 cannot safely include a path component.

I think we should be more specific here. The Base URL should contain only the "scheme" and "authority" (commonly known as "host and port") portions of a [RFC 3986 URI](https://tools.ietf.org/html/rfc3986).

We will likely change this to allow a path component in v6, as discussed in #55.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
+# Twirp Wire Protocol
+
+This document defines the Twirp wire protocol over HTTP.
```
spenczar: Can we include somewhere that this describes version 5 of Twirp?

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+* **Interface** is the proto `service` name for an API. For example,
+  `CalendarService`.
+
+* **Method** is the proto `rpc` name for an API method. For example,
+  `CreateEvent`.
+
+### Requests
+
+**Request ::= Request-Headers Request-Body**
```
spenczar: I'm not sure what this line is getting us. Can we delete it?

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+Twirp always uses HTTP POST method to send requests, because it
+closely matches the semantics of RPC methods.
+
+The **Request-Headers** are normal HTTP headers. The Twirp wire
+protocol uses the following headers.
+
+* **Authorization** header is often used to pass user credentials
```
spenczar: This is unused by Twirp, please remove.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+  from the client to the server, such as OAuth access token or
+  JWT token.
+
+* **Content-Type** header indicates the proto message encoding, which
+  should be one of "application/x-protobuf", "application/json". The
```
spenczar: We accept `application/protobuf`, not `application/x-protobuf`.

If we change that in v6, we can say so in a v6 spec, but we should describe the current state of things in this PR.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+  server uses this value to decide how to parse the request body,
+  and encode the response body.
+
+* **User-Agent** header indicates the client application and its
```
spenczar: This is unused by Twirp, please remove.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+  runtime environment. While the server should not use this
+  information for request processing, this header is heavily used
+  for analytics and troubleshooting purposes.
+
+* **RPC-Timeout** header indicates the client-specified request
```
spenczar: This is unused by Twirp, please remove.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+  server should use a pre-configured timeout value, by default it
+  should be 10 seconds.
+
+The **Request-Body** is the encoded request message, contained in the
+HTTP request body. The encoding is specified by the `Content-Type`
+header.
+
+### Responses
+
+**Response ::= Response-Headers Response-Body**
```
spenczar: I think we can drop this line, too.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+The **Response-Headers** are just normal HTTP response headers. The
+Twirp wire protocol uses the following headers.
+
+* **Content-Type** The value should be either "application/x-protobuf"
```
spenczar: As above, should be `application/protobuf`

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Example
+
+The following example shows a simple Echo API definition and its
+corresponding wire payloads.
+
+The example assumes the server base URL is "https://example.com".
+
+```proto
+package twirp;
```
spenczar: Missing `syntax = "proto3";` at the top of this proto file.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+  string message;
+}
+```
+
+**Proto Request**
+
+```
+POST /twirp.Echo/Hello HTTP/1.1
+Host: example.com
+Content-Type: application/x-protobuf
```
spenczar: should be `application/protobuf`

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+Content-Length: 15
+
+<encoded HelloRequest>
```
spenczar: Should we try to represent the encoded protobuf here in some way? I think that might be valuable.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+The following example shows a simple Echo API definition and its
+corresponding wire payloads.
+
+The example assumes the server base URL is "https://example.com".
+
+```proto
+package twirp;
+
+service Echo {
```
spenczar: Let's change this to the classic Twirp example, [`Haberdasher`](https://github.com/twitchtv/twirp/blob/master/example/service.proto) :)

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Errors
+
+If an error occurs when the server processes a request, the server
+must return an error payload as the response message, and correctly
+set the HTTP status code. Please see
+[`google.rpc.Code`](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto)
+on how to map typical server errors to HTTP status codes.
+
+### Timeout errors
```
spenczar: This is not true of Twirp, please remove this section.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+For a single request, there is a client-specified timeout and a
+server-configured timeout. If a request misses the server-configured
+timeout, the server must return a `503` error. If a request misses
+the client-specified timeout that it is shorter then the server-
+configured timeout, the server must return a `504` error.
+This allows more accurate measurement of the server availability.
+
+### Network errors
+
+If a client fails to reach the server due to network errors, the
```
spenczar: I don't think we should specify this. I don't think client libraries should be exposing HTTP status codes, just Twirp error codes.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+
+{"message":"Hello, World!"}
+```
+
+## Errors
+
+If an error occurs when the server processes a request, the server
+must return an error payload as the response message, and correctly
+set the HTTP status code. Please see
+[`google.rpc.Code`](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto)
```
spenczar: We don't strictly follow `google.rpc.Code`.  We use our own error codes, heavily based on those but with some modifications: the codes are a string on the wire for human readability, and we add the `bad_route` error code. See https://github.com/twitchtv/twirp/blob/master/errors.go#L138-L275.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+Content-Type: application/json
+Content-Length: 27
+
+{"message":"Hello, World!"}
+```
+
+## Errors
+
+If an error occurs when the server processes a request, the server
+must return an error payload as the response message, and correctly
```
spenczar: I think we should describe Twirp's error payload in detail here.

Twirp error responses are always JSON-encoded (*regardless* of the request's Content-Type), with a corresponding `Content-Type: application/json` header. This ensures that they are human-readable in any setting.

They are a JSON object with three keys:

 - **code**: One of the Twirp error codes as a string.
 - **msg**: A human-readable message describing the error as a string.
 - **meta**: An object with string keys and values holding arbitrary additional metadata describing the error.

For example:
```json
{
    "code": "permission_denied",
    "msg": "thou shall not pass",
    "meta": {
        "target": "Balrog"
    }
}
```

See https://github.com/twitchtv/twirp/wiki/Errors.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
+# Twirp Wire Protocol
+
+This document defines the Twirp wire protocol over HTTP.
+
+## Conventions
+
+The requirement level keywords "MUST", "MUST NOT", "REQUIRED",
```
spenczar: None of these keywords are used in this document. Let's drop this paragraph.

PROTOCOL.md
```diff
@@ -0,0 +1,193 @@
...
+"SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY",
+and "OPTIONAL" used in this document are to be interpreted as
+described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+The grammar rules used in this document are using [ABNF
```
spenczar: We only use ABNF in one place where it's useful, which is in describing the URL. Let's delete this paragraph and just say something like `In [ABNF syntax](https://tools.ietf.org/html/rfc5234), Twirp's URLs follow this format: ...`

Then, we can get rid of this `Conventions` section altogether, which is good.

PROTOCOL.md
mocax: Echo is the most classic example for network APIs used in many places. As a protocol spec, we should use the example most developers are already familiar with.

For programming languages, everyone chooses hello-world. I think we should just follow the historical tradition.

PROTOCOL.md
mocax: This is a public spec. I believe 99% developers don't know or care about proto encoding format. Let's wait and see. We can always add it if many people ask for it.

mocax:
Addressed most comments.

PROTOCOL.md
```diff
@@ -0,0 +1,174 @@
...
+
+message HelloResponse {
+  string message;
+}
+```
+
+**Proto Request**
+
+```
+POST /twirp.Echo/Hello HTTP/1.1
```
spenczar: The URL here should be `/twirp/twirp.Echo/Hello`. Same for JSON.

The stuttering of 'twirp' here is not ideal. Could you change the package to something like `package example.echoer`?

mocax:
Done.

PR #73, opened 2018-01-31
reduce duplicate go generator code

jmank88:
This PR proposes simplifying some generator code by combining two sets of mostly duplicate JSON/Protobuf code.  The generated code remains unchanged.


spenczar:
Hmmm, I think this change makes things a little more complex, not less. I think the generators' dirt-simple structure right now makes them very easy to understand. Abstraction here doesn't seem worth it.

While abstraction here would make it easier to make changes to both JSON and Protobuf methods at the same time, I don't quite think that's worth it in this case.

I'm willing to hear arguments, but closing this for now.

PR #76, opened 2018-02-09
Remove all uses of stdlib log

spenczar:
*Issue #, if available:*

#4 

*Description of changes:*

This removes the last references to the stdlib `log` package in generated code.

Three cases change - in all these cases, we used to log:
 1. If we get an error on req.Body.Close(), we now call the Error hook if no other errors occurred.
 2. If we get an error on resp.Body.Write(), we now call the Error hook.
 3. If we get an error on resp.Body.Write() while writing a serialized error back to the client, we ignore the error.

cc @cep21


protoc-gen-twirp/generator.go
```diff
@@ -966,14 +968,22 @@ func (t *twirp) generateServerJSONMethod(service *descriptor.ServiceDescriptorPr
 	methName := stringutils.CamelCase(method.GetName())
 	t.P(`func (s *`, servStruct, `) serve`, methName, `JSON(ctx `, t.pkgs["context"], `.Context, resp `, t.pkgs["http"], `.ResponseWriter, req *`, t.pkgs["http"], `.Request) {`)
 	t.P(`  var err error`)
+	t.P()
+	t.P(`  defer func() {`)
+	t.P(`    closeErr := req.Body.Close()`)
```
rhysh: The net/http.Server will take care of closing the body. I don't see the benefit of closing it ourselves here, and the extra code and path to calling the Error hook makes these functions more complex.

From `go doc http.Request.Body`,

> The Server will close the request body. The ServeHTTP Handler does not need to.

I'd prefer that you delete this section (and the matching one in `generateServerProtobufMethod`), but it's up to you.

protoc-gen-twirp/generator.go
```diff
@@ -1018,7 +1028,8 @@ func (t *twirp) generateServerJSONMethod(service *descriptor.ServiceDescriptorPr
 	t.P(`  resp.Header().Set("Content-Type", "application/json")`)
 	t.P(`  resp.WriteHeader(`, t.pkgs["http"], `.StatusOK)`)
 	t.P(`  if _, err = resp.Write(buf.Bytes()); err != nil {`)
-	t.P(`    `, t.pkgs["log"], `.Printf("errored while writing response to client, but already sent response status code to 200: %s", err)`)
+	t.P(`    err = wrapErr(err, "failed to write response to client")`)
+	t.P(`    callError(ctx, s.hooks,`, t.pkgs["twirp"], `.InternalErrorWith(err))`)
```
rhysh: Using an "Internal" error to call the hook seems weird. Write errors here are caused by things like clients that have reset the connection .. that seems like an "external" error :)

I read through the error codes we have and didn't see one that fits well.

If we know that it's because the client disconnected, it might be appropriate to return a "Canceled" error .. but if we do that, then maybe we should check against the Context value (which net/http should have canceled)? Is that complexity worthwhile?

We could make a rough attempt to classify the error into "Canceled", maybe "DeadlineExceeded", and maybe others .. and use "Unknown" for anything we don't understand?

protoc-gen-twirp/generator.go
```diff
@@ -352,9 +350,23 @@ func (t *twirp) generateUtils() {
...
+	t.P(`    // We have three options here. We could log the error, call the Error`)
+	t.P(`    // hook, or just silently ignore the error.`)
+	t.P(`    //`)
+	t.P(`    // Logging is unacceptable because we don't have a user-controlled `)
+	t.P(`    // logger; writing out to stderr without permission is too rude.`)
+	t.P(`    //`)
+	t.P(`    // Calling the Error hook would confuse users: it would mean the Error`)
+	t.P(`    // hook got called twice for one request, which is likely to lead to`)
+	t.P(`    // duplicated log messages and metrics, no matter how well we document`)
+	t.P(`    // the behavior.`)
```
rhysh: This seems accurate. Error-while-erroring seems fine to ignore.

protoc-gen-twirp/generator.go
spenczar: Yeah, "Unknown" might be more appropriate here for now. I don't think I want to get into the weeds of diagnosing whether it was a context cancellation. That seems hard to get right, and I'd rather conservatively label this as unknown.

I've been also mulling a `"TransportFailure"` error code, to describe all these sorts of low-level problems. I don't think this is the time to add it, but this might be a good use for it.

protoc-gen-twirp/generator.go
spenczar: Agreed, will do.

spenczar:
@rhysh, please take another look here.

PR #77, opened 2018-02-12
docs/spec_v5_errors.md

marioizquierdo:
The current `docs/errors.md` page was only useful for Go services. This PR defines Twirp errors on a wire protocol level. This will be useful for services and clients in other languages.


docs/spec_v5.md
```diff
@@ -151,30 +148,20 @@ Content-Length: 27
...
+  "code": "permission_denied",
+  "msg": "thou shall not pass",
+  "meta": {
+    "target": "Balrog"
+  }
 }
 ```
 
-For more information, see https://github.com/twitchtv/twirp/wiki/Errors.
+For more information see the [Errors Spec](spec_v5_errors.md).
```
spenczar: I think I'd like to have the whole spec live in one document. Could you move the list of error codes and the status code mapping into this file, and delete `doc/spec_v5_errors.md`?

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+{"code":"internal","msg":"Something went wrong"}
+```
+
+### Twirp Error Codes
+
+The possible values for the Error "code" are intentionaly similar to
+[gRPC status codes](https://godoc.org/google.golang.org/grpc/codes):
+	
+* **canceled**: The operation was cancelled (typically by the caller).
+* **unknown**: Unknown error. For example when handling errors raised by APIs
```
spenczar: Rephrase this to complete sentences: `"An unknown error occurred. For example, this can be used when handling errors raised by APIs that do not return any error information."`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Twirp Error Codes
+
+The possible values for the Error "code" are intentionaly similar to
+[gRPC status codes](https://godoc.org/google.golang.org/grpc/codes):
+	
+* **canceled**: The operation was cancelled (typically by the caller).
+* **unknown**: Unknown error. For example when handling errors raised by APIs 
+  that do not return enough error information.
+* **invalid_argument**: the client specified an invalid argument. It
```
spenczar: Capitalize `The` to be consistent.

Change `It` to `This`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+  indicates arguments that are problematic regardless of the state of the
```
spenczar: change `problematic` to `invalid`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+  system (i.e. a malformed file name, required argument, number out of range, etc.).
+* **deadline_exceeded**: Operation expired before completion. For operations
```
spenczar: Complete sentences: `An operation expire before completion`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+  operation has completed successfully (timeout).
+* **not_found**: Some requested entity was not found.
+* **bad_route**: The requested URL path wasn't routable to a Twirp
+  service and method. This is returned by the generated server, and usually
+  shouldn't be returned by applications. Instead, applications should use
+  "not_found" or "unimplemented".
+* **already_exists**: An attempt to create an entity failed because one
+  already exists.
+* **permission_denied**: The caller does not have permission to execute
+  the specified operation. It must not be used if the caller cannot be
```
spenczar: Change `It` to `This`.

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+  For example, seeking or reading past end of a paginated collection.
+  Unlike "invalid_argument", this error indicates a problem that may be fixed if
+  the system state changes (i.e. adding more items to the collection).
+  There is a fair bit of overlap between "failed_precondition" and "out_of_range".
+  We recommend using "out_of_range" (the more specific error) when it applies so
+  that callers who are iterating through a space can easily look for an
+  "out_of_range" error to detect when they are done.
+* **unimplemented**: The operation is not implemented or not
+  supported/enabled in this service.
+* **internal**: When some invariants expected by the underlying system
```
spenczar: Complete sentences: remove leading `When`.

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+  have been broken. In other words, something bad happened in the library or
+  backend service. Twirp specific issues like wire and serialization problems
+  are also reported as "internal" errors.
+* **unavailable**: The service is currently unavailable. This is most
+  likely a transient condition and may be corrected by retrying with a
+  backoff.
+* **data_loss**: Unrecoverable data loss or corruption.
```
spenczar: Complete sentences: `... or corruption has occurred.`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+
+{"code":"internal","msg":"Something went wrong"}
+```
+
+### Twirp Error Codes
+
+The possible values for the Error "code" are intentionaly similar to
+[gRPC status codes](https://godoc.org/google.golang.org/grpc/codes):
+	
+* **canceled**: The operation was cancelled (typically by the caller).
```
spenczar: Remove `(typically by the caller)`.

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+Example Error Response:
+```
+HTTP/1.1 500 Internal Server Error
+Content-Type: application/json
+Content-Length: 48
+
+{"code":"internal","msg":"Something went wrong"}
+```
+
+### Twirp Error Codes
```
spenczar: I think this would be easier to read as a table. Could you try something like this:

Code | HTTP Status Code | Description
----|----|----
canceled | 408 | The operation was cancelled (typically by the caller).
unknown | 500| Unknown error. For example when handling errors raised by APIs that do not return enough error information.

etc

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+"unimplemented"       | 501 Not Implemented
+"internal"            | 500 Internal Server Error
+"unavailable"         | 503 Service Unavailable
+"dataloss"            | 500 Internal Server Error
+```
+
+### Non-Twirp Errors from Intermediary Proxies
+
+It is also possible for Twirp Clients to receive non-200 responses with invalid 
+Twirp Errors (non JSON, or with invalid keys), this responses may be
```
spenczar: typo: `this responses`

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+generated by proxy middleware or network issues. Twirp Clients should build a
+valid Twirp error depending on the HTTP status of those invalid responses:
+
+```
+HTTP status code           Twirp Error Code
```
spenczar: I don't think we should specify this at the specification level. I don't see when it would need to be interoperable across languages - I think we can get away with saying that the choice here is an implementation detail.

It's like how HTTP doesn't specify that "status code" that you should report to clients if a TCP connection breaks. Similarly, we don't need to specify what clients should present if the remote end _isn't_ speaking Twirp. We should only cover what to do if it _is_ speaking Twirp.

However, this is still useful to document somewhere, maybe in `docs/errors.md`.

docs/spec_v5_errors.md
```diff
@@ -0,0 +1,173 @@
...
+503 Service Unavailable  | "unavailable"
+504 Gateway Timeout      | "unavailable"
+... other                | "unknown"
+```
+
+Additional metadata should be added to make it easy to identify intermediary errors:
+* `"http_error_from_intermediary": "true"`
+* `"status_code": string` (original status code on the HTTP response, e.g. `"500"`).
+* `"body": string` (original non-Twirp error response as string).
+* `"location": url-string` (only on 3xx reponses, matching the `Location` header).
```
spenczar: Definitely don't think this should be in the spec, but definitely think it should be in the docs somewhere that we do this in Go.

docs/errors.md
```diff
@@ -10,12 +10,15 @@ error)`.
 Twirp clients always return errors that can be cast to `twirp.Error`. Even
 transport-level errors will be `twirp.Error`s.
 
-Twirp server implementations can return regular errors, but if they do, they
+Twirp server implementations can return regular non-twirp errors too, but those
```
spenczar: Remove `non-twirp` here.

docs/spec_v5.md
marioizquierdo: Can we do that as a next step? (it will be an easy followup PR). This way we keep this PR clean with the comments in place.

docs/spec_v5_errors.md
marioizquierdo: Yeah I though about this, but I don't see why Twirp clients need to be inconsistent across languages and why should we let them re-discover the same issues over an over. Twirp doesn't only cover service and wire, but also auto-generated clients, so it made sense to include some notions about how to handle non-twirp responses.

In any case, it is true that this part makes the spec a little noisy, and I was using "should" vocabulary instead of "must". Maybe we should write a spec for Twirp Clients? Or a doc with recommendations? Or as you said, just including this in `docs/errors.md` could already work as a point of reference to other languages.

spenczar: Twirp's specification should be as small as possible. If this doesn't need to be in the spec, it shouldn't be.

Let's put this in a separate document like `docs/errors.md` to explain how it works in Go, but leave it up to language-specific implementers. It should be allowed to do this in a way that makes sense in their language.

docs/spec_v5.md
spenczar: No, I'd like to do this in one go. It's better if future readers can see the changes to the spec (and discussions about it) by looking at the git history of just the one file.

I don't think it should be too noisy - it will mostly be adding text, not editing any, right?

marioizquierdo: Yes that's right. Will merge in next commit 👍

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
-## Errors
 
-Twirp error responses are always JSON-encoded, regardless of
-the request's Content-Type, with a corresponding
-`Content-Type: application/json` header. This ensures that
-the errors are human-readable in any setting.
+## Errors
 
-Twirp errors are a JSON object with three keys:
+Erros are non-200 JSON responses with the keys:
```
spenczar: Please restore this to the original paragraph from before.

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
+    "target": "Balrog"
+  }
+}
 ```
+
+### Error Responses are JSON
+
+Twirp error responses are always JSON-encoded, the response Content-Type
+header is always `application/json`, regardless of the request's 
+Content-Type (protbuf requests also respond with json errors).
```
spenczar: I think this summary belongs at the top, as I mentioned at L154

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
+
+```
+HTTP/1.1 500 Internal Server Error
+Content-Type: application/json
+Content-Length: 48
+
+{"code":"internal","msg":"Something went wrong"}
+```
+
+### Twirp Error Codes
```
spenczar: Retitle to just `Error Codes`

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
+
+The possible values for the Error "code" are intentionaly similar to
```
spenczar: Revise this section's text to this:

"Twirp errors always include an error code. This code is represented as a string and must be one of a fixed set of codes, listed in the table below. Each code has an associated HTTP Status Code. When a server responds with the given error code, it must set the corresponding HTTP Status Code for the response."

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
+  "msg": "please use a smaller size",
+  "meta": {
+    "argument": "size",
+    "max_allowed": "1000"
+  }
 }
 ```
 
-For more information, see https://github.com/twitchtv/twirp/wiki/Errors.
+Error metadata can only have string values. This is to simplify error parsing by clients.
```
spenczar: I don't think we need this paragraph. Let's put it in `docs/errors.md` instead.

docs/spec_v5.md
```diff
@@ -151,30 +148,91 @@ Content-Length: 27
...
+| unimplemented       | 501 | The operation is not implemented or not supported/enabled in this service.
+| internal            | 500 | When some invariants expected by the underlying system have been broken. In other words, something bad happened in the library or backend service. Twirp specific issues like wire and serialization problems are also reported as "internal" errors.
+| unavailable         | 503 | The service is currently unavailable. This is most likely a transient condition and may be corrected by retrying with a backoff.
+| dataloss            | 500 | The operation resulted in unrecoverable data loss or corruption.
+
+
+### Metadata
```
spenczar: Remove this section - it's already covered above. The spec should be terse.

docs/errors.md
```diff
@@ -35,35 +38,56 @@ type Error interface {
...
+| DataLoss           | dataloss            | 500 Internal Server Error
+
+For more information on each code, see the [Errors Spec](spec_v5.md).
+
+### HTTP Errors from Intermediary Proxies
+
+It is also possible for Twirp Clients to receive non-200 responses with invalid 
+Twirp Errors (non JSON, or with invalid keys) that may be generated by 
+intermediary proxies like a load balancer. Twirp clients return valid Twirp 
+errors depending on the HTTP status of the invalid responses:
```
spenczar: Rephrase to 

"It is also possible for Twirp Clients to receive HTTP responses with non-200 status codes but without an expected error message. For example, proxies or load balancers might return a "503 Service Temporarily Unavailable" body, which cannot be deserialized into a Twirp error.

In these cases, generated Go clients will return `twirp.Error`s with a code which depends upon the HTTP status of the invalid response:"

PR #80, opened 2018-02-14
Add draft of v6 specification

spenczar:
This is a prerelease draft to explain planned changes coming in v6. The only change
from v5 (so far) is a change of the URL scheme Twirp uses.

*Issue #, if available:*

Relates to #55, but doesn't implement it yet.

*Description of changes:*

Here's `diff PROTOCOL.md PROTOCOL_V6_DRAFT.md`:
```diff
< # Twirp Wire Protocol
---
> # EXPERIMENTAL Twirp v6 Wire Protocol
3,4c3,31
< This document defines the Twirp wire protocol over HTTP. The
< current protocol version is v5.
---
> This document is a draft of a new version ("v6") of the Twirp wire
> protocol over HTTP. This draft is EXPERIMENTAL and subject to change.
> v6 of Twirp is not released or considered stable; instead, this
> document lists planned changes which may be available in prerelease
> distributions.
>
> ## Changes from v5
>
> ### URL scheme
>
> In [v5](./PROTOCOL.md), URLs followed this format:
>
> **URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
>
> In v6, the "/twirp/" prefix is removed:
>
> **URL ::= Base-URL "/" [ Package "." ] Service "/" Method**
>
> The "/twirp/" prefix is removed for three reasons:
>
>  - Trademark concerns: some very large organizations don't want to
>    take any legal risks and are concerned that "twirp" could become
>    trademarked.
>  - Feels like advertising: To some users, putting "twirp" in all your
>    routes feels like it's just supposed to pump Twirp's brand, and
>    provides no value back to users.
>  - Homophonous with "twerp": In some Very Serious settings (like
>    government websites), it's not okay that "twirp" sounds like
>    "twerp", which means something like "insignificant pest."
28c55
< **URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
---
> **URL ::= Base-URL "/" [ Package "." ] Service "/" Method**
109c136
< POST /twirp/example.echoer.Echo/Hello HTTP/1.1
---
> POST /example.echoer.Echo/Hello HTTP/1.1
120c147
< POST /twirp/example.echoer.Echo/Hello HTTP/1.1
---
> POST /example.echoer.Echo/Hello HTTP/1.1
```


spenczar:
cc @mocax, @rhysh, @marioizquierdo, @Xe.

docs/spec_v6.md
```diff
@@ -0,0 +1,259 @@
...
+
+ - Trademark concerns: some very large organizations don't want to
+   take any legal risks and are concerned that "twirp" could become
+   trademarked.
+ - Feels like advertising: To some users, putting "twirp" in all your
+   routes feels like it's just supposed to pump Twirp's brand, and
+   provides no value back to users.
+ - Homophonous with "twerp": In some Very Serious settings (like
+   government websites), it's not okay that "twirp" sounds like
+   "twerp", which means something like "insignificant pest."
```
marioizquierdo: This is a concern only for whoever is updating from v5 to v6, I think it is better to keep the v6 (latest) spec clear without this. It may be enough with adding this info in the Release Notes for v6, or if you want to keep this info in the spec is better to add it in the v5 spec (as a notice that it changed after v6).

docs/spec_v6.md
```diff
@@ -0,0 +1,259 @@
...
+
+To use Twirp, developers first define their APIs using proto files,
+then use Twirp tools to generate the client and the server libraries.
+The generated libraries implement the Twirp wire protocol, using the
+standard HTTP library provided by the programming language runtime or
+the operating system. Once the client and the server are implemented,
+the client can communicate with the server by making RPC calls.
+
+The Twirp wire protocol supports both binary and JSON encodings of
+proto messages, and works with any HTTP client and any HTTP version.
```
marioizquierdo: duplicated overview

docs/spec_v6.md
```diff
@@ -0,0 +1,259 @@
...
+
+The Twirp wire protocol uses HTTP URLs to specify the RPC
+endpoints on the server for sending the requests. Such direct mapping
+makes the request routing simple and efficient. The Twirp URLs have
+the following components.
+
+* **Base-URL** is the virtual location of a Twirp API server, which is
+  typically published via API documentation or service discovery.
+  Currently, it should only contain URL `scheme` and `authority`. For
+  example, "https://example.com".
```
marioizquierdo: in v6, the Base-URL should allow a path prefix as well, for example "http://mydomain.com/twirp", specially important to ease backwards compatibility with v5 services.

docs/spec_v6.md
```diff
@@ -0,0 +1,259 @@
...
+* **Package** is the proto `package` name for an API, which is often
+  considered as an API version. For example,
+  `example.calendar.v1`. This component is omitted if the API
+  definition doesn't have a package name.
+* **Service** is the proto `service` name for an API. For example,
+  `CalendarService`.
+* **Method** is the proto `rpc` name for an API method. For example,
+  `CreateEvent`.
+
```
marioizquierdo: Would it make sense to add an example? It may be useful, specially because Base-URL contains "/", and Package contains "." on it (same symbols used as separators). The example should make it visually clear that Twirp URLs need to be parsed from right to left (extract Method first, then Service, then optional Package, and the rest is the Base-URL).

docs/spec_v6.md
```diff
@@ -0,0 +1,259 @@
...
+| unimplemented       | 501 | The operation is not implemented or not supported/enabled in this service.
+| internal            | 500 | When some invariants expected by the underlying system have been broken. In other words, something bad happened in the library or backend service. Twirp specific issues like wire and serialization problems are also reported as "internal" errors.
+| unavailable         | 503 | The service is currently unavailable. This is most likely a transient condition and may be corrected by retrying with a backoff.
+| dataloss            | 500 | The operation resulted in unrecoverable data loss or corruption.
```
marioizquierdo: I think you need an extra breakline at the end to make sure that the table is properly redered

PROTOCOL_V6_DRAFT.md
```diff
@@ -0,0 +1,201 @@
...
+
+In [v5](./PROTOCOL.md), URLs followed this format:
+
+**URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
+
+In v6, the "/twirp/" prefix is removed:
+
+**URL ::= Base-URL "/" [ Package "." ] Service "/" Method**
+
+The "/twirp/" prefix is removed for three reasons:
```
mocax: The "/twirp/" prefix is not really removed. It can be part of the Base-URL if the service provider wants to keep it. It is just no longer part of the spec.

PROTOCOL_V6_DRAFT.md
```diff
@@ -0,0 +1,201 @@
...
+
+### URL scheme
+
+In [v5](./PROTOCOL.md), URLs followed this format:
+
+**URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
+
+In v6, the "/twirp/" prefix is removed:
+
+**URL ::= Base-URL "/" [ Package "." ] Service "/" Method**
```
mocax: Please remove the first "/" and make it part of Base-URL. The less client logic, the better.

PROTOCOL_V6_DRAFT.md
```diff
@@ -0,0 +1,201 @@
...
+
+The Twirp wire protocol supports both binary and JSON encodings of
+proto messages, and works with any HTTP client and any HTTP version.
+
+### URLs
+
+In [ABNF syntax](https://tools.ietf.org/html/rfc5234), Twirp's URLs
+have the following format:
+
+**URL ::= Base-URL "/" [ Package "." ] Service "/" Method**
```
mocax: This URL schema is identical to gRPC, which seems confusing. At least it would conflict with existing logging, monitoring, routing, load balancing systems. While I am not saying gRPC should own this URL schema, it is better not to be identical between Twirp and gRPC.

docs/spec_v6.md
mocax: We should remove this restriction. The Base-URL can be any valid URL. There is no need to add any restriction. For example, why do we need to rule out port?

docs/spec_v6.md
mocax: There is no need to parse the URL. The server registers a handler against a URL, and all it needs to do is URL lookup using a hash table.

docs/spec_v6.md
spenczar: A URL "authority" includes a port, but you're right that we should allow path prefixes.

I don't think we should allow any valid URL. We should not permit query parameters or fragments. I think this should be come  "it should only contain a URL `scheme`, `authority`, and `path`, as defined in RFC 3986 Section 3."

docs/spec_v6.md
spenczar: Yeah, I'll move this sort of thing into a changelog doc that will go unpublished until we release v6.

docs/spec_v6.md
spenczar: There's an example further down that I think should be enough.

I agree that URLs don't need to be parsed like that necessarily. For example, if the user mounts the Twirp server under a path prefix, we could say they should trim the path prefix from request URLs before handing the request off to Twirp in the Go implementation. In any case, this is an implementation-specific concern.

PROTOCOL_V6_DRAFT.md
Xe: I thought the schema was supposed to be `URL ::= Base-URL "/" [ Package "/" ] Service "/" Method`

PROTOCOL_V6_DRAFT.md
spenczar: Right, we should definitely allow path prefixes.

PROTOCOL_V6_DRAFT.md
spenczar: I think we need the leading "/". If it's in the Base-URL, we need to say that the URL must not have an empty path. Otherwise, we'd allow a Base-URL of `https://example.com` which would lead to this:
`https://example.comtwirp.example.Haberdasher/MakeHat`

which seems clearly invalid.

peterhellberg: Wouldn't you use the `ResolveReference` method on `*url.URL` in order to support both cases?

Eg. https://play.golang.org/p/Zj6ehRCuZ_G

spenczar: @peterhellberg That might be a good way to make the API nice for Go users, but this question is on what the spec should be, which is a language-independent question. We're trying to work out what the BNF for the URL scheme should look like, which is probably a bit pedantic, but important to get right.

spenczar:
All concerns addressed, I believe - please take another look, y'all.

docs/spec_changelog.md
```diff
@@ -0,0 +1,43 @@
...
+### URL scheme
+
+In [v5](./PROTOCOL.md), URLs followed this format:
+
+```bnf
+**URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
+```
+
+Version 6 changes this format to remove the mandatory `"/twirp"` prefix, and
+changes `Package` and `Service` to be delimited with a `/` instead of a `.`
```
rhysh: Packages will still contain `.` characters; we're not changing the internal delimiter. How about this phrasing:

"and changes the separator between `Package` and `Service` to be a `/` instead of a `.` character"

docs/spec_changelog.md
```diff
@@ -0,0 +1,43 @@
...
+
+```bnf
+**URL ::= Base-URL "/" [ Package "/" ] Service "/" Method**
+```
+
+Also, `Base-URL` can now contain a path component - in other words, it's legal
+to set any prefix you like.
+
+If you loved the old `/twirp` prefix, you can still use it by using a base URL
+that ends with `/twirp`. You're no longer forced into use it, however.
```
rhysh: Users can keep that prefix, but 1) they'll need to remove it before passing the request to the Twirp generated code, and 2) the rest of the URI path won't be the same as v5 because of the `.`->`/` change (unless the package is unset).

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
+---
+id: "spec_v6"
+title: "Twirp Wire Protocol"
+sidebar_label: "Version 6 (Proposed Draft)"
+---
+
+This document is a draft of a new version ("v6") of the Twirp wire
+protocol over HTTP. This draft is EXPERIMENTAL and subject to change.
+v6 of Twirp is not released or considered stable; instead, this
```
rhysh: "not yet released ...", unless you want something like "even-numbered releases are unstable"

docs/spec_v6.md
rhysh: Needing to parse right-to-left is worrisome. That will be hard to implement consistently and to explain concisely. It sounds like we're not quite heading in that direction .. but if we are, we should stop.

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+
+message HelloResponse {
+  string message;
+}
+```
+
+**Proto Request**
+
+```
+POST /userprefix/example.echoer/Echo/Hello HTTP/1.1
```
rhysh: Using `/` as a delimiter forces us to deal with cases of overlap. Consider the following two:

One has no package declaration, and describes a service called "Foo" with a method "Hello". It gets requests like `POST /userprefix/Foo/Hello`.

The other has a package declaration of "package Foo;" .. not matching the normal lower-case pattern, but a package declaration nonetheless. It describes some service, maybe "Bar" with a method "Goodbye". It gets requests like `POST /userprefix/Foo/Bar/Goodbye`.

How does a user route requests to those two services? Do they need to count the slashes? Peeking into the later parts of the URI path and only partially understanding them is an invitation for maintenance and security problems down the road.

docs/spec_v6.md
spenczar: @rhysh, certainly agreed, but I don't think that's what we're headed to. In Go, users would do something like this:
```go
mux.Handle("/userdefined" + HaberdasherPathPrefix, http.StripPrefix("/userdefined", svc))
```

docs/spec_changelog.md
spenczar: Right, I'll clarify this.

docs/spec_v6.md
spenczar: Sure, not yet released - I'm just trying to explain that it's a draft.

docs/spec_v6.md
spenczar: Oof, this is an extremely good point, and might torpedo the `/` idea here. I'm going to remove the `.` -> `/` change from this commit.

Let's discuss in #86.

docs/spec_changelog.md
```diff
@@ -0,0 +1,43 @@
...
+**URL ::= Base-URL "/twirp/" [ Package "." ] Service "/" Method**
+```
+
+Version 6 changes this format to remove the mandatory `"/twirp"` prefix, and
+changes `Package` and `Service` to be delimited with a `/` instead of a `.`
+character:
+
+```bnf
+**URL ::= Base-URL "/" [ Package "/" ] Service "/" Method**
+```
```
marioizquierdo: I'm not sure if there was already a discussion about using "." vs "/" for the package separator, but I am definitely not convinced that doing the `. -> /` change is a good thing. Let me list the advantages that I can see with each one:

Advantages of "/" separator:
  * Looks nicer when described like this: `Base/Package/Service/Method`.
  * It is different from gRPC routes. This was mentioned before as a pro, but I don't clearly see the advantage of this; only the very trained eye will be able to tell the difference at a glance, and I don't see the advantage of keeping a Twirp and a gRPC service both in the same Base-URL, only differentiated by the "." vs "/" syntax.
  * If there is no package or the package has no dots, and the Base-URL also has no dots, then the resulting URL will not have dots. Maybe this is easier to recognize by some middleware?

Advantages of "." separator:
  * Somewhat backwards-compatible. With the "/", service implementations will need to explicitly handle both new and old paths. With the "." this is not entirely true, because even if the service didn't support old routes, the user could still add the "/twirp" prefix on the Base-URL to deal with compatibility.
  * Since packages are optional, the "." makes it easier to tell if a URL has a package (`blahblah.com/api/myserviceapi.CoolAPI/CreateStuff`) or not (`blahblah.com/api/CoolAPI/CreateStuff`). 
 * Avoids service path conflicts. For example, given the services `Foo1` with package `package` and name `Foo`, and `Foo2` with no package and the same name `Foo`. With the "." a URL like `http://fooservice.com/package/Foo/Method` is clearly mapping to `Foo2` with the Base-URL `http://fooservice.com/package`, but with the "/" the same URL is ambiguous. This is probably an edge case because in most cases the user controls the Base-URL, but it is still possible.

I may be missing some other point (please comment!), but the advantages of the "/" separator look a little weak to me.

docs/spec_v6.md
marioizquierdo: 😅 I retire the sentence "need to be parsed from right to left". Sorry that was misleading. The URLs can be parsed in multiple ways, one of them is by starting from the right, but it could also be removing the Base-URL first and then using a regexp, etc. What I mean is that one example will make it easier to understand the things that need to be consider when implementing a Twirp router, or just to understand the routes themselves.

mocax:
For protos without package, we don't really need to care. It is extremely
rare and should be highly discouraged. I don't think it would cause any
real problem.

docs/spec_changelog.md
spenczar: Excellent points, and I agree that we under-designed this. I think PR feedback is a tough place to do design discussion, though. Could you chime in on #86? You can just paste in what you have here :)

marioizquierdo: Ok I'll add this points to #86. Now I realize they are very similar to what @rhysh was talking about, but presented in a different order. It may help to add clarity to design decisions.

spenczar:
@mocax, I don't think I agree. It may be extremely rare within Google, but it is not so rare in the wild. We need to make a spec sturdy enough that this sort of thing doesn't cause problems for our users.

We could require `package` be set (making it an even breakier change!) or we could back away and keep `.` as the delimiter. Let's discuss that in #86. For now, I'm going to change this PR to only remove the `/twirp` prefix.

docs/spec_v6.md
lwc: Yeah this would be be great - we use path prefixes to route between our services so twirp is a non-starter for us atm.

mocax:
Given all the troubles for change . to / as the separator, I agree we
should drop the idea. It doesn't seem to have enough real value for such a
breaking change. I think we can move forward without it.

spenczar:
@marioizquierdo and @rhysh, please take another look. I think I've addressed all concerns here.

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+  "code": "permission_denied",
+  "msg": "Thou shall not pass",
+  "meta": {
+    "target": "Balrog",
+    "power": "999"
+  }
+}
+```
+
+### Error Codes
```
Xe: I very much like this section and how it's formatted

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+
+Twirp errors always include an error code. This code is represented
+as a string and must be one of a fixed set of codes, listed in the
+table below. Each code has an associated HTTP Status Code. When a
+server responds with the given error code, it must set the
+corresponding HTTP Status Code for the response.
+
+| Twirp Error Code    | HTTP Status | Description
+| ------------------- | ----------- | -----------
+| canceled            | 408 | The operation was cancelled.
```
Xe: Are these error codes sent as strings over the wire?

spenczar: Yep, they are.

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+syntax = "proto3";
+
+package example.echoer;
+
+service Echo {
+  rpc Hello(HelloRequest) returns (HelloResponse);
+}
+
+message HelloRequest {
+  string message;
```
rhysh: `string message = 1;`

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+}
+
+message HelloResponse {
+  string message;
```
rhysh: `string message = 1;`

docs/spec_v6.md
```diff
@@ -0,0 +1,220 @@
...
+`Content-Type: application/json` header. This ensures that
+the errors are human-readable in any setting.
+
+Twirp errors are a JSON object with the keys:
+
+* **code**: One of the Twirp error codes as a string.
+* **msg**: A human-readable message describing the error
+  as a string.
+* **meta**: (optional) An object with string values holding
+  arbitrary additional metadata describing the error.
```
rhysh: Use of "arbitrary" here makes it sound like they don't need to be strings .. even though "string values" makes it clear that they need to be.

Consider s/metadata/string-formatted metadata/

docs/spec_v6.md
rhysh: The error code is `canceled`, which is the spelling the Go project uses. Consider s/cancelled/canceled/ in the description for consistency :)

PR #83, opened 2018-02-15
Set Accept header in generated client requests

### LINKED ISSUE #81: Generated clients should set Accept header to right Content-Type
I am road testing twirp with amazon api gateway and lambda, this service requires the `Accept` header as this triggers binary encoding of the response for `application/protobuf`.

I am happy to do a PR for this.

wolfeidau:
fixes #81

Added Accept header to JSON and protobuf clients and ran make.

### DIFF
--- a/protoc-gen-twirp/generator.go
+++ b/protoc-gen-twirp/generator.go
@@ -418,6 +418,7 @@ func (t *twirp) generateUtils() {
 	t.P(`  if customHeader := getCustomHTTPReqHeaders(ctx); customHeader != nil {`)
 	t.P(`    req.Header = customHeader`)
 	t.P(`  }`)
+	t.P(`  req.Header.Set("Accept", contentType)`)
 	t.P(`  req.Header.Set("Content-Type", contentType)`)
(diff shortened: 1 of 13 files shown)


protoc-gen-twirp/generator.go
```diff
@@ -649,6 +651,8 @@ func (t *twirp) generateUtils() {
 	t.P(`  if err != nil {`)
 	t.P(`    return clientError("could not build request", err)`)
 	t.P(`  }`)
+	t.P(`  req.Header.Set("Accept", "application/json")`)
```
spenczar: If the user has set their own `Accept` header (through `WithHTTPRequestHeaders`), we will be overriding it.

We need to either document this in `WithHTTPRequestHeaders` or rearrange things so that the `WithHTTPRequestHeaders` value wins.

I would prefer rearranging things. That way we aren't breaking compatibility in any way.

wolfeidau:
I have updated this PR as requested, thanks for the hint on `WithHTTPRequestHeaders` my fix was obviously pretty crap when compared with the existing header configuration.

Issue #86, opened 2018-02-20
Proposal: Use <package>/<service> instead of <package>.<service> in URLs


spenczar:
This obviously overlaps with #55, but the discussion there focused on removal of the `/twirp` prefix. I'd like a new issue to concentrate discussion on just the `.` to `/` switch.

The switch from a `.` to delimit package and service towards `/` as delimiter is really consequential and needs its own clearer discussion. Let's work out whether it's a good idea.

### Path-based routing in other infrastructure

Lots of intermediary proxies, load balancers, caches, and so on will naturally split a URL on `/` characters. Using the `/` character will make it easier to configure those tools. For example, a load balancer could more simply route requests to the right backend if it knew that the second path component is the service name.

### Long-term cruft in generated APIs

We are removing the `/twirp` prefix, and instead permitting user-defined prefixes.

Migrating a server from v5 to v6 is easy if this is the only change. To support v5 clients, you do this:
```go
mux := http.NewServeMux()
mux.Handle(HaberdasherPathPrefix, habderdasher)
mux.Handle("/twirp" + HaberdasherPathPrefix, http.StripPrefix("/twirp", haberdasher))
```

If we change the `.` to `/` too, though, we need to export an additional path prefix const. We'll need these:
```go
const (
  HaberdasherPathPrefix = "/twirp.example/Haberdasher/"
  HaberdasherV5PathPrefix = "/twirp/twirp.example.Haberdasher/"
)
```
And the user will need to use the right one in their mux:

```go
mux := http.NewServeMux()
mux.Handle(HaberdasherPathPrefix, habderdasher)
mux.Handle(HaberdasherV5PathPrefix, haberdasher)
```

Generated code would need to route these v5 requests correctly in addition to the new v6 requests. We'd need backwards compatibility shims in generated code, probably forever, and we would need to keep the extra exported const around forever.

### Route ambiguity when package is unset

Proto files need not set `package`. What do we do in those cases? @rhysh outlined the way this could go wrong. For illustration, suppose you had these two services:
```protobuf
package Foo

service Bar {
  rpc Handle(Req) returns (Resp);
}
```
and 
```protobuf
// package is never specified

service Foo {
  rpc Bar(Req) returns (Resp);
}
```

What are the routes? If we leave `.`, the routes are:
```
  /Foo.Bar/Handle -> Bar service's Handle method
  /Foo/Bar        -> Foo service's Bar method
```

If we change them to `/`, though, they are:
```
[...]

PR #91, opened 2018-03-15
Remove /twirp prefix from routes

spenczar:
*Issue #, if available:*

#86 

*Description of changes:*

This change implements the plan discussed in #86: It removes the /twirp prefix from routes.

This would be merged into a 'v6_prerelease' branch, not master.

### DIFF
--- a/clientcompat/internal/clientcompat/clientcompat.twirp.go
+++ b/clientcompat/internal/clientcompat/clientcompat.twirp.go
@@ -157,7 +157,7 @@ func (s *compatServiceServer) writeError(ctx context.Context, resp http.Response
 // CompatServicePathPrefix is used for all URL paths on a twirp CompatService server.
 // Requests are always: POST CompatServicePathPrefix/method
 // It can be used in an HTTP mux to route twirp requests along with non-twirp requests on other routes.
-const CompatServicePathPrefix = "/twirp/twirp.clientcompat.CompatService/"
+const CompatServicePathPrefix = "/twirp.clientcompat.CompatService/"
 
 func (s *compatServiceServer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
@@ -180,10 +180,10 @@ func (s *compatServiceServer) ServeHTTP(resp http.ResponseWriter, req *http.Requ
 	}
 
 	switch req.URL.Path {
-	case "/twirp/twirp.clientcompat.CompatService/Method":
+	case "/twirp.clientcompat.CompatService/Method":
 		s.serveMethod(ctx, resp, req)
 		return
-	case "/twirp/twirp.clientcompat.CompatService/NoopMethod":
+	case "/twirp.clientcompat.CompatService/NoopMethod":
 		s.serveNoopMethod(ctx, resp, req)
 		return
 	default:
--- a/clientcompat/pycompat/clientcompat_pb2_twirp.py
+++ b/clientcompat/pycompat/clientcompat_pb2_twirp.py
@@ -49,7 +49,7 @@ class CompatServiceClient(object):
 
     def __make_request(self, body, full_method):
         req = Request(
-            url=self.__target + "/twirp" + full_method,
+            url=self.__target + full_method,
             data=body,
             headers={"Content-Type": "application/protobuf"},
         )
--- a/example/service.twirp.go
+++ b/example/service.twirp.go
@@ -137,7 +137,7 @@ func (s *haberdasherServer) writeError(ctx context.Context, resp http.ResponseWr
 // HaberdasherPathPrefix is used for all URL paths on a twirp Haberdasher server.
 // Requests are always: POST HaberdasherPathPrefix/method
 // It can be used in an HTTP mux to route twirp requests along with non-twirp requests on other routes.
-const HaberdasherPathPrefix = "/twirp/twitch.twirp.example.Haberdasher/"
+const HaberdasherPathPrefix = "/twitch.twirp.example.Haberdasher/"
 
 func (s *haberdasherServer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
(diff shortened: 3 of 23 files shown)


clientcompat/internal/clientcompat/clientcompat.twirp.go
```diff
@@ -180,10 +180,10 @@ func (s *compatServiceServer) ServeHTTP(resp http.ResponseWriter, req *http.Requ
 	}
 
 	switch req.URL.Path {
-	case "/twirp/twirp.clientcompat.CompatService/Method":
+	case "/twirp.clientcompat.CompatService/Method":
 		s.serveMethod(ctx, resp, req)
 		return
-	case "/twirp/twirp.clientcompat.CompatService/NoopMethod":
+	case "/twirp.clientcompat.CompatService/NoopMethod":
```
marioizquierdo: Just removing the prefix is not enough. The base-URL now can have path components. For example, to upgrade this service while keeping compatibility, before was mounted on "http://localhost:8000" but now it should be mounted on "http://localhost:8000/twirp". The routing should check for `req.URL.Path` suffix, not exact match.

clientcompat/pycompat/clientcompat_pb2_twirp.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, server_address):
 
     def __make_request(self, body, full_method):
         req = Request(
-            url=self.__target + "/twirp" + full_method,
+            url=self.__target + full_method,
```
marioizquierdo: Clients should be fine with this. The only thing required to keep backwards compatibility would be to change the base-URL (__target) to include the "/twirp" prefix 👍

example/service.twirp.go
```diff
@@ -137,7 +137,7 @@ func (s *haberdasherServer) writeError(ctx context.Context, resp http.ResponseWr
 // HaberdasherPathPrefix is used for all URL paths on a twirp Haberdasher server.
 // Requests are always: POST HaberdasherPathPrefix/method
 // It can be used in an HTTP mux to route twirp requests along with non-twirp requests on other routes.
-const HaberdasherPathPrefix = "/twirp/twitch.twirp.example.Haberdasher/"
+const HaberdasherPathPrefix = "/twitch.twirp.example.Haberdasher/"
```
marioizquierdo: I wonder if it would be better to add a new constant name to make sure that the compiler tells the developer that there's something that needs to change after the upgrade. Since the "PathPrefix" now is just the service full name with `/`, maybe it would make sense to change the constant to this?
```go
const HaberdasherFullName = "twitch.twirp.example.Haberdasher"
```

protoc-gen-twirp/generator.go
```diff
@@ -881,7 +881,7 @@ func (t *twirp) generateServer(file *descriptor.FileDescriptorProto, service *de
 // service. It includes a trailing slash. (for example
 // "/twirp/twitch.example.Haberdasher/").
 func pathPrefix(file *descriptor.FileDescriptorProto, service *descriptor.ServiceDescriptorProto) string {
-	return fmt.Sprintf("/twirp/%s/", fullServiceName(file, service))
+	return fmt.Sprintf("/%s/", fullServiceName(file, service))
```
marioizquierdo: need to update comment

rhysh:
How much risk is there in keeping the symbol names the same while changing their meaning? `NewHaberdasherServer` now expects to get requests without a leading `/twirp` and will respond with a "bad_route" error if it gets requests from v5 clients. Users who upgrade from v5 to v6 without reading and understanding the patch notes will break.

We could make `NewHaberdasherServer` accept requests both with and without the `/twirp` prefix, and ship an `HTTPClient` ("Doer") implementation that removes the `/twirp` prefix (since the client behavior wouldn't be able to immediately change). It would stick around forever.

We could change the symbol names to `NewHaberdasherServerV6` and `NewHaberdasherProtobufClientV6` so it's clear which ones don't use the `/twirp` prefix. We could keep the old constructors, at least until we release v7.0.0. Or we could add the `...V6` constructors in v5.X.0 and remove the old ones in v6.0.0.

On issue tracking, it looks like this code implements the proposal at #55, minus the `.` vs `/` question described in #86.

clientcompat/internal/clientcompat/clientcompat.twirp.go
spenczar: Of course, I am not sure how I forgot that in this first pass. Will fix.

spenczar:
I am very much OK with keeping the same symbol names for the PathPrefix const and for the constructor because this change increments the major version number. This is what semantic versioning is for. Cluttering the API is not worth it.

Serving requests both with and without the `/twirp` prefix, on the other hand, might be a worthwhile thing that helps people upgrade; I'm more inclined to do that.1

clientcompat/internal/clientcompat/clientcompat.twirp.go
spenczar: Oh, I _do_ remember how: we don't want a suffix match; we expect users to use [`net/http.StripPrefix`](https://golang.org/pkg/net/http/#StripPrefix) when setting up their server if they want to use a prefix.

spenczar:
@rhysh @marioizquierdo, please take another look.

clientcompat/pycompat/clientcompat_pb2_twirp.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, server_address):
```
marioizquierdo: for readability it could make sense to rename `server_address` to `base_url`, with an example like `http://localhost:8080/twirp`

clientcompat/internal/clientcompat/clientcompat.twirp.go
marioizquierdo: Ah ok, but then we will have to be careful documenting this in the docs. Please make sure that the Haberdasher example shows how to properly mount the service in the "/twirp" prefix, with a NOTE for older versions.

Issue #94, opened 2018-03-27
Context setters don't work when imported packages' vendor directories are not flattened


kminehart:
I have an external service called "Metadata", which is a twirp server.

Here's a small test app I wrote to try to send it an HTTP header, basically following the example in the wiki verbatim.

```go
package main

import (
	"context"
	"log"
	"net/http"
	"os"

	pb "git.example.io/rune/rune/metadata/rpc"
	"github.com/twitchtv/twirp"
)

var (
	twirpContext = context.Background()
)

func init() {
	header := make(http.Header)
	log.Println(header)
	header.Set("Example-Authorization", "testing")
	var err error
	twirpContext, err = twirp.WithHTTPRequestHeaders(twirpContext, header)
	if err != nil {
		log.Panicln("Error attaching headers to twirp context: ", err.Error())
	}
}

func main() {
	log.Println(twirpContext)
	exampleClient := pb.NewMetadataProtobufClient(os.Getenv("RUNE_METADATA_ADDRESS"), &http.Client{})
	exampleClient.Get(twirpContext, &pb.GetMetadataRequest{})
}

```

Log output:

```
2018/03/27 13:17:23 map[]
2018/03/27 13:17:23 context.Background.WithValue(5, http.Header{"Example-Authorization":[]string{"testing"}})
```

Here is the request being logged when Go / twirp is used:

```
...metadata.Metadata/Get] Headers: map[Content-Type:[application/protobuf] Twirp-Version:[v5.2.0] Accept-Encoding:[gzip] X-Envoy-Internal:[true] X-Envoy-Expected-Rq-Timeout-Ms:[5000] User-Agent:[Go-http-client/1.1] Content-Length:[0] X-Forwarded-Proto:[https] X-Request-Id:[eb43abc0-bbe9-4d7e-9f37-f709e82949e1] X-Ambassador-Calltype:[extauth-request] X-Forwarded-For:[10.24.5.119]]
```
You'll notice `Twirp-Version` is there, but not `Example-Auth`.

But when I use curl:

```
curl -L -v -X POST -H "content-type: application/json" -d '{"test":"req"}' -H "Example-Authorization: testing" https://api.example.io/metadata/v1/twirp/example.metadata.Metadata/Get
```

```
[...]

PR #95, opened 2018-03-28
Fix Twirp clients not sending HTTP headers when server uses vendored twirp

kminehart:
*Issue #, if available:* #94

*Description of changes:*

1. I added a few basic tests to test some functions in `/context.go`; while this was for my own sanity, I figured if they exist they might as well be put into the PR.

2. in `/internal/contextkeys/keys.go`, I removed the custom type:

```Go
type contextKey int
```

as it was creating a type mismatch when importing a Twirp client from a repository with a vendored version of Twirp.

As @zikes explained to me,

> `context` value lookup says there's a difference between `github.com/twitchtv/twirp/internal/contextkeys.contextKey(5)` and `git.example.io/example/project/vendor/github.com/twitchtv/twirp/internal/contextkeys.contextKey(5)`, which is what `contextkey.RequestHeaderKey` resolves to.

So, by removing this custom type and simply using `int`, this essentially solves that problem.

3. I added go-cmp as a dependency. It's not 100% necessary, but it does make life a lot easier when writing tests compared to `reflect.DeepEqual`. If this is a problem I can make due without it.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -16,8 +16,25 @@
 [[projects]]
   branch = "master"
   name = "github.com/golang/protobuf"
-  packages = ["jsonpb","proto","protoc-gen-go/descriptor","protoc-gen-go/plugin","ptypes/struct"]
+  packages = [
+    "jsonpb",
+    "proto",
+    "protoc-gen-go/descriptor",
+    "protoc-gen-go/plugin",
+    "ptypes/struct"
+  ]
   revision = "1e59b77b52bf8e4b449a57e6f79f21226d571845"
+
+[[projects]]
+  name = "github.com/google/go-cmp"
+  packages = [
+    "cmp",
+    "cmp/internal/diff",
+    "cmp/internal/function",
+    "cmp/internal/value"
+  ]
+  revision = "3af367b6b30c263d47e8895973edcca9a49cf029"
+  version = "v0.2.0"
 
 [[projects]]
   name = "github.com/pkg/errors"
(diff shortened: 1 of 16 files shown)


spenczar:
Hi, sorry for the delay here - I've been on vacation for a week.

Thanks for making this change! I don't think we should do this, though.

Switching to an `int` type for those consts means that we are storing values in the `Context` with just literal `0`, `1`, `2`, `3`... values. The chance of collision with other packages, or with user-written code, is _way_ higher.

This is called out by the [documentation in the `context.Context` package](https://golang.org/pkg/context/), and is a requirement for safely using contexts:

        // A key identifies a specific value in a Context. Functions that wish
        // to store values in Context typically allocate a key in a global
        // variable then use that key as the argument to context.WithValue and
        // Context.Value. A key can be any type that supports equality;
        // packages should define keys as an unexported type to avoid
        // collisions.
        //
        // Packages that define a Context key should provide type-safe accessors
        // for the values stored using that key:
        //
        // 	// Package user defines a User type that's stored in Contexts.
        // 	package user
        //
        // 	import "context"
        //
        // 	// User is the type of value stored in the Contexts.
        // 	type User struct {...}
        //
        // 	// key is an unexported type for keys defined in this package.
        // 	// This prevents collisions with keys defined in other packages.
        // 	type key int
        //
        // 	// userKey is the key for user.User values in Contexts. It is
        // 	// unexported; clients use user.NewContext and user.FromContext
        // 	// instead of using this key directly.
        // 	var userKey key

So, we need to use a custom type here; I don't think we can go with this solution. I don't think the problem is one for Twirp to solve, really, but I'll put my thoughts on that on #94.

Issue #13, opened 2018-01-17
Handle the standard M= and import_prefix= golang plugin flags


peter-edge:
Example: https://github.com/yarpc/yarpc-go/blob/dev/internal/protoplugin/runner.go#L50

Will be needed when request/response types are outside of the package of the service, or if generating the Twirp handlers to a different golang package.

Nice work by the way!

spenczar:
Right, when we wrote the generator we skipped this part because it doesn't seem to be documented anywhere very well, so it seemed too hard to get right.

Could you write a bit more about what these things are used for, and what their precise meaning is?

spenczar:
@peter-edge ping! i'm inclined to close this without further info :)

jonathaningram:
@spenczar I just came across this when trying to organise my protos in a package or two :(

Hopefully this can help:

Say my proto files are something like:

```proto
// ./root-path/rpc/empty.proto
// Note in this case by "rpc" I don't mean Twirp's suggested folder structure, I just mean that the Empty message is related to general RPC calls.
syntax = "proto3";
package rpc;

// Empty is a message used when there is no value returned from a call.
message Empty{}
```

```proto
// ./root-path/my-domain/my-domain.proto
syntax = "proto3";
package mydomain;

import "rpc/empty.proto";

service MyDomainService {
  rpc Noop(rpc.Empty) returns (rpc.Empty);
}
```

And say I have a `generate.sh` file that looks something like:

```sh
#!/usr/bin/env bash

ROOT_PATH="something"

PKG_PREFIX="github.com/my-company/my-repo"

PKG_MAP=""
PKG_MAP="$PKG_MAP,Mrpc/empty.proto=$PKG_PREFIX/some-path/proto/rpc"

protoc --proto_path $PROTO_PATH --go_out=$PKG_MAP:$PROTO_PATH $PROTO_PATH/rpc/*.proto
protoc --proto_path $PROTO_PATH --twirp_out=$PROTO_PATH --go_out=$PKG_MAP:$PROTO_PATH $PROTO_PATH/my-domain/*.proto
```

[...]

PR #102, opened 2018-04-18
Add support for import map and import prefix

jonathaningram:
*Issue #, if available:*

Refs #13.

*Description of changes:*

Provides the same support that `protoc-gen-go` supports for import mapping and import prefix.

I haven't written any docs on this...


spenczar:
As we discussed in #13, I think it's better to encourage people to use `option go_package` to indicate how things should be imported.

protoc-gen-twirp/generator.go
```diff
@@ -61,14 +64,43 @@ func newGenerator() *twirp {
...
+		}
+	}
+
+	t.importMap = make(map[string]string)
+	for k, v := range param {
+		switch k {
+		case "import_prefix":
+			t.importPrefix = v
+		default:
+			if len(k) > 0 && k[0] == 'M' {
```
spenczar: Ugh, this is a terrible API for users... but it _is_ what `github.com/golang/protobuf/protoc-gen-go` does. I suppose we should support it.

But can we also provide a less-insane format? I don't know exactly what, but here's an idea:
```
protoc --twirp_out="go_package@rpc/empty.proto=github.com/twitchtv/rpc/empty,go_package@something/else.proto=golang.org/x/net/notreal.proto:.
```

protoc-gen-twirp/generator.go
```diff
@@ -61,14 +64,43 @@ func newGenerator() *twirp {
...
 		output:              bytes.NewBuffer(nil),
 	}
 
 	return t
 }
 
+// commandLineParameters breaks the comma-separated list of key=value pairs
+// in the parameter (a member of the request protobuf) into a key/value map.
+// It then sets file name mappings defined by those entries.
+func (t *twirp) commandLineParameters(parameter string) {
```
spenczar: Instead of making this a method on `*twirp`, could you make this a function at the top level that returns a struct that has the fields that come from the command line? I think that would make it easier to write unit tests and also make it easier to document and keep track of which parameters are set in the command line like this. 

I'm picturing the call site as something like this: 
```
params, err := parseParams(in.GetParameters())
if err != nil {
  gen.Fail("invalid parameters passed to --twirp_out: %v", err)
}
t.importMap = params.importMap
t.importPrefix = params.importPrefix
```

protoc-gen-twirp/generator.go
```diff
@@ -61,14 +64,43 @@ func newGenerator() *twirp {
...
+	param := make(map[string]string)
+	for _, p := range strings.Split(parameter, ",") {
+		if i := strings.Index(p, "="); i < 0 {
+			param[p] = ""
```
spenczar: Eh, I think we can error on this. Let's only support `k=v`-style parameters.

jonathaningram:
@spenczar cheers, will address.

jonathaningram:
Quick update: I think I've addressed all feedback except the docs. I went with your suggestion for `go_package@`. A better option is not really apparent to me at the moment...

jonathaningram:
@spenczar thinking about the prefix to use for this, `go_package` is probably confusing as it conflicts with the option used in the proto file and also doesn't really say what it's doing.

What about `go_import_mapping`? So the option looks like:

```
go_import_mapping@rpc/empty.proto=github.com/user/rpc
```

1. It has a `go` prefix since this is for Go things.
2. The parameter prefix maps back to the Go variable name we actually use in the generator which is named `importMap`
3. In this way, the protoc-gen-go `M` actually becomes a shorthand flag despite missing the `@`.

jonathaningram:
@spenczar ok that's enough PR spam for you. Could you please review? I've added a doc file and also decided to change to `go_import_mapping` per the above. I think it makes much more sense.

PR #109, opened 2018-05-30
Add doc page with a cURL example

### LINKED ISSUE #22: Add curl example using protobuf format to wiki
The wiki has an example of how to curl using the json format here:
https://github.com/twitchtv/twirp/wiki/HTTP-Routing-and-Serialization#making-requests-on-the-command-line-with-curl

It would be useful I think if this section also contained an example of how to curl using protobuf, which is possible if the user has `protoc` installed, which is likely since it's required for building the autogenerated code.

e.g.
```
echo "inches:10" \
	| protoc --proto_path=$GOPATH/src --encode twirp.example.haberdasher.Size ./rpc/haberdasher/service.proto \
	| curl -s --request POST --header "Content-Type: application/protobuf" --data-binary @- http://localhost:8080/twirp/twirp.example.haberdasher.Haberdasher/MakeHat \
	| protoc --proto_path=$GOPATH/src --decode twirp.example.haberdasher.Hat ./rpc/haberdasher/service.proto
```

I'd open a PR to add it to docs, but I don't see a way of doing that since the docs are in the github wiki and not in the repo.

leighmcculloch:
What
===
Add doc page with a cURL exaple.

Why
===
cURLing is an important part of being able to debug a service. It's
worth providing examples using JSON and Protobufs as it can be daunting
figuring out how to do both correctly.

Resolves #22

### DIFF
--- a/docs/curl.md
+++ b/docs/curl.md
@@ -1,0 +1,80 @@
+---
+id: "curl"
+title: "cURL"
+sidebar_label: "cURL"
+---
+
+Twirp allows you to cURL your service with either Protobuf or JSON.
+
+## Example
+
+A cURL request to the Haberdasher example `MakeHat` RPC with the following request and reply could be executed as Protobuf or JSON with the snippets further below.:
+
+Request proto:
+```
+message Size {
+   int32 inches = 1;
+}
+```
+
+Reply proto:
+
+```
+message Hat {
+  int32 inches = 1;
+  string color = 2;
+  string name = 3;
+}
+```
+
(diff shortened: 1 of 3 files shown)


spenczar:
Looks good, thanks. Sorry I have to be a stickler on the legal side, but can you confirm that your contribution is made under the terms of the Apache 2.0 license?

leighmcculloch:
Yup, confirmed this contribution is made under the terms of the Apache 2.0 license.

Issue #70, opened 2018-01-29
Proposal: Go streaming API


spenczar:
#3 lays out some ideas for how Twirp could support streaming RPCs on the wire. Equally important is how we support them in generated code.

This issue is for designing the generated Go code. When we've landed on a proposal that seems pretty good, I'll edit this top text with what we've got, and then we can look at implementing it.

Let's use this service to discuss what things would look like:

```proto
syntax = "proto3";

service Streamer {
  rpc Transaction(Req) returns (Resp);
  rpc Upload(stream Req) returns (Resp);
  rpc Download(Req) returns (stream Resp);
  rpc Bidirectional(stream Req) returns (stream Resp);
}

message Req {}
message Resp {}
```

spenczar:
Here's one way to do this.

The generated interface would use `<Msg>Stream` types:

```go
type Streamer interface {
  Transaction(context.Context, *Req) returns (*Resp, error)
  Upload(context.Context, ReqStream) returns (*Resp, error)
  Download(context.Context, *Req) returns (RespStream, error)
  Bidirectional(context.Context, ReqStream) returns (RespStream, error)
}
```

These Stream types are an interface:

```go
type ReqStream interface {
  Next(context.Context) (*Req, error)
}
```

#### As the recipient:
You call `Next` to get another message from the stream. If the stream has been closed by the sender, `Next` returns a nil `*Req`, and an error that describes the shutdown. This could either be a `var ErrStreamComplete`, indicating a "healthy" termination of the stream, or an error that indicates a problem.

To shut down the stream early as the recipient, cancel the context. Otherwise, you must call `Next` repeatedly until it returns `nil, <error>`. Failure to do one of these two things will leave the HTTP connection hanging open.

#### As the sender:
It's your responsibility to provide an implementation of this interface. The generated Twirp code will be calling `Next`, then serializing the objects it gets and shipping them to the recipient. It will call `Next` repeatedly until `Next` returns `nil, <error>`, so an implementation must promise to eventually return `nil, <error>` or it will leak an HTTP connection.

We will provide a constructor for a simple implementation of the interface for convenience. It might look like this, but I'm not attached to this API and it probably needs work:

```go
type reqStreamSender struct {
  ch <- chan *Req
}

func NewReqStream(ch chan *Req) *reqStreamSender {
  return &reqStreamSender{ch: ch}
}

func (s *reqStreamSender) Next(ctx context.Context) (*Req, error) {
  select {
    case <- ctx.Done():
[...]

Issue #3, opened 2018-01-16
Add streaming support?


jacksontj:
I think this is a great addition for RPC for Go (as well as other languages!). The main thing that I expect will limit adoption is the lack of streaming support. There are mechanisms to do streaming on http/1.x (a great example is the [grpc-gateway stuff](https://github.com/grpc-ecosystem/grpc-gateway)) where they effectively just use chunked encoding with blobs to do the streaming.

spenczar:
Yeah, let's get this discussion going. I agree that this is an important feature.

One important constraint: we need to continue to work on HTTP 1.1. I *think* that means we can't ever support full-duplex, bidirectional streaming, but I'd be very happy to be proven wrong.

Could you (and others) write a bit about the use cases for streaming RPCs? At Twitch, we haven't really needed any, so I'd like to have concrete users to build for to guide the design of the user API. The underlying implementation seems pretty clear, but I want to make sure the generated code feels nice to use.

marwan-at-work:
My use case is usually file uploads from one backend service to another. I've written a CI/CD where I have multiple containers build different docker images and stream data from all these containers to a central backend that tars all of them and uploads them as a github release asset. Those could be many megabytes and obviously had to be chunk'd and sent out as streams. RPC streams were perfect for my use case : )

tommie:
I don't have any needs for Twirp, but I do use gRPC streaming.

In all cases so far, I've never need bi-directional on the same RPC. The one place I could think of is if you have a single RPC that sends state deltas and receives state deltas. But in that case, you're really looking for classic long-polling e.g. Websockets, not the RPC style of doing things. You can always invent an identifier on top of the RPC mechanism that would allow concurrency safe pub-sub on two uni-directional streams.

So if the only concern is support for bi-directionality, I'd say start with mutually exclusive directions, WLOG. That also cuts the question of full duplex short. I mean, on the wire, HTTP/1.1 could definitely support full duplex streaming, but I assume the problem is that the XHR API requires a request to be finished before reading the response and is hence practically unusable. (Is it more common for HTTP server APIs to support FD BD streaming? Thinking about the Go HTTP library got me to "yes".)

thelazyfox:
We should also consider websockets.  Client support is pretty widely available these days and we'd get bidirectional streaming support out of the box.

[...]

PR #112, opened 2018-06-06
WIP: Version 6

spenczar:
*Issue #, if available:*
#3 
#55
#70 
#80

*Description of changes:*

This PR holds all the v6 changes, so they can be merged into master at once. While working on these features, they will be merged into the v6_prerelease branch to make them available for adventurous users.


jtvjorge:
For anyone still interested, this is on hold until at least March of 2019. Thanks for your interest!

rmasp:
> For anyone still interested, this is on hold until at least March of 2019. Thanks for your interest!

Interested. And it's March. Questions/comments:

1. There's been no updates for about 6 months. Is this release still happening?
2. I only want the routing changes (wish that was the original protocol), not any streaming stuff. Why delay and complicate v6 with more moving parts?
3. `git checkout master && git merge v6_prerelease` looks pretty clean and simple. How far out on a limb would I be if I fork and track my own until this gets done?

Great "batteries included" stuff here!

fi0:
I wish the custom prefix #55 could be added in v5, instead of waiting for v6. Or streaming stuff can be a v7?

spenczar:
Yes, as I stated in #70, I think streaming might be the wrong next step for Twirp. That would simplify this release considerably, changing it down to just being about paths.

>  How far out on a limb would I be if I fork and track my own until this gets done?

@rmasp, I'm not sure exactly. The current branch _just_ removes the `/twirp` prefix, but I think I might want to do something a little more sophisticated to make the migration process easier. That might come with a new API, or something - not exactly sure, but I'm back to thinking about it much more actively.

Thanks for waiting, and sorry about the long delay on this. The reason for the delay is mostly that Twirp works pretty well today, and these routing changes will take some careful work to make the transition smooth. I think they're basically just cosmetic, too.

rmasp:
> The current branch just removes the /twirp prefix, but I think I might want to do something a little more sophisticated to make the migration process easier

And that's all I want/need. Again, for those of us exposing our Twirp APIs to 3rd parties, the /twirp is an issue. However, I branched `master`, merged in `v6_prerelease` and everything looks good. And since I'm building anew, migration is a non-issue for me.

That said, why not make the lead path part configurable at protoc code-gen time, with nil or empty string leaving that path piece out entirely? Default would be `/twirp`, so legacy code would not be affected?

spenczar:
Right, I'd just like to avoid configuration at code-gen time if possible. Code generators are often buried in tool chains, so passing through options to them can be hard for people with any sophistication in their builds. An option in the `.proto` file might work, but that would require providing an importable definition, which adds its own headaches... maybe worth it, but it's a challenge.

rmasp:
Seems like the approach I like puts the onus of toolchain config on those if us who don't like the default. It's a no-op for anyone happy with the `/twirp` default. Just sayin'.

spenczar:
Yep, but if I can avoid adding that complexity for _anyone_, I'd prefer that. Good point though, you've moved my thinking on this a bit.

rmasp:
Thanks for listening! Fwiw, I'm using twirp in a pipeline with gqlgen, which has a bunch of code gen configurability in a yml: https://gqlgen.com/config/

Imho as twirp sees wider adoption you'll get requests for a more configurable settings. Look at Django, Rails, ... I know it's a fine line between simplicity and configurability. Way to many tools (esp in the golang space) have more dials and buttons than Spock's science station. Have to design for the "just right porridge" and common case defaults.

Bankq:
Shall we merge newer releases back to `v6_prerelease` branch in the meantime?

rmasp:
Fwiw, next week I'm moving my servers (and clients) to the v6_prerelease branch because of product requirements that include no /twirp in path. I'd really like to see v6 become mainline.

z0mbie42:
Hi,
Any news regarding a merge date ?

Also, keep the good work, twrip is awesome!

z0mbie42:
Hi @spenczar ,
Any news regarding a merge date ?

Also, thank you, twrip is awesome!

marioizquierdo:
Sorry, this PR is not going to be merged. The [protocol v6](https://twitchtv.github.io/twirp/docs/spec_v6.html) proposal was closed because both issues included in the release could not reach consensus (streaming: https://github.com/twitchtv/twirp/issues/3, and updating the path: https://github.com/twitchtv/twirp/issues/55).

This doesn't mean that the features will never be implemented in some form, this only means that they don't depend on each other for v6.

PR #114, opened 2018-06-16
gofmt

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

https://blog.golang.org/go-fmt-your-code

---
> I made this PR with a project going on over at https://github.com/rotblauer/gofmt-att, and it's definitely a work in progress. So if I got something wrong, or this is annoying at all, please file an issue over there and we'll sort it out.

### DIFF
--- a/_tools/src/github.com/gogo/protobuf/protoc-gen-gogo/grpc/grpc.go
+++ b/_tools/src/github.com/gogo/protobuf/protoc-gen-gogo/grpc/grpc.go
@@ -138,7 +138,7 @@ func (g *grpc) GenerateImports(file *generator.FileDescriptor) {
 
 // reservedClientName records whether a client name is reserved on the client side.
 var reservedClientName = map[string]bool{
-// TODO: do we need any in gRPC?
+	// TODO: do we need any in gRPC?
 }
 
 func unexport(s string) string { return strings.ToLower(s[:1]) + s[1:] }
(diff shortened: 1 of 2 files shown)


rhysh:
Thank you for sending the patch, but there are two reasons I think we won't merge it. I'll cross-link in the gofmt-att repo, as requested.

First, the code in the _tools directory is copied in automatically by [retool](https://github.com/twitchtv/retool). That means that this repo isn't the right place to make changes to that code—any changes should happen in the upstream source. Retool places the code in a directory that doesn't match `...` (it starts with an underscore), so it's surprising that code there is inspected by scripts like gofmt-att.

Not wanting to maintain patches to tools like that is the main reason I think we won't merge this change.

Second, the code in question almost certainly is already go-formatted .. according to the rules in a different version of Go. Some of the rules [changed in Go 1.10](https://golang.org/doc/go1.10#gofmt), and trusting a particular version of Go to see if code is formatted will lead to unnecessary churn.

whilei:
Roger that @rhysh -- thanks for the feedback. Since running my last batch of PRs I've learned about this `_` pre-/suffixing convention and have added it to my blacklist to avoid outcomes like this.

whilei:
Also that's a really important link on the gofmt changes -- thanks for that.

spenczar:
@rhysh covered the reasons for closing this well.

Issue #123, opened 2018-08-09
Generated code will not compile if the service defined in protobuf definition has a method defined with the same name.


darren-west:
Hi,

While evaluating twirp I came across a small issue. I created a simple Protobuf service definition that echoes a message. I decided to call the service Echo with a method of Echo for simplicity. I found that the generated code did not compile. 

Protobuf example below:

```protobuf
syntax = "proto3";

package twitch.twirp.echo;
option go_package = "api";

service Echo{
    rpc Echo(Message) returns (Message) {}
}

message Message {
    string Value = 1;
}

```

Although this is probably a rare case, I looked into it and its because it uses an embedded struct in the generated server struct. 

I will be creating a PR with the fix. 

Thanks,
Darren

spenczar:
Fixed in #124, release v5.4.2.

PR #124, opened 2018-08-09
Fix bug with code generation of service with a method of the same name.

darren-west:
Issue #123 

The generated api.twirp.go file will not compile if a service contains a method with the same name as the services name. This is because the generated code uses an embedded struct.

This fix uses the embedded structs name when making a call to the method.

I can't see any unit tests for the generated code. If I get time and don't need to refactor too much I may add one. 


darren-west:
Sorry I missed the tests when I looked quickly. 

I've now added a test for compilation.

darren-west:
I've pushed the changes. 

The Go generated files look good. There is a small change in the python files though. Not sure if they were out of date?

PR #126, opened 2018-09-14
Improve support for go_package option in .proto files

dbudworth:
Previously, specifying the option go_package caused the generator
to write a file to the fully qualified path
This would create the full directory tree

this doesn't really work as it adds, to your project root dir, the full github.com/blah/...
structure

This change does two things
adds support for the `paths=source_relative` like the google protoc plugin does
which ensures we write to whatever directory the .proto file lives in

secondly, it makes sure to use whatever go_package says when importing the module.
It was generating imports with relative dirs from the project root

With these two changes, you can use twirp with go 1.11's modules now

### DIFF
--- a/protoc-gen-twirp/generator.go
+++ b/protoc-gen-twirp/generator.go
@@ -48,7 +48,7 @@ type twirp struct {
 	importMap    map[string]string // Mapping from .proto file name to import path.
 
 	// Package output:
-	paths string // instruction on where to write output files
+	sourceRelativePaths bool // instruction on where to write output files
 
 	// Package naming:
 	genPkgName          string // Name of the package that we're generating
@@ -85,7 +85,7 @@ func (t *twirp) Generate(in *plugin.CodeGeneratorRequest) *plugin.CodeGeneratorR
 
 	t.genFiles = gen.FilesToGenerate(in)
 
-	t.paths = params.paths
+	t.sourceRelativePaths = params.paths == "source_relative"
 
 	// Collect information on types.
 	t.reg = typemap.New(in.ProtoFile)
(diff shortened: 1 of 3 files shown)


spenczar:
This sounds great, I totally agree that it is needed. I’m on vacation but should be able to take a closer look on Monday or Tuesday. Thanks very much.

protoc-gen-twirp/generator.go
```diff
@@ -47,6 +47,9 @@ type twirp struct {
 	importPrefix string            // String to prefix to imported package file names.
 	importMap    map[string]string // Mapping from .proto file name to import path.
 
+	// Package output:
+	paths string // instruction on where to write output files
```
spenczar: Let's make this `sourceRelativePaths bool` instead of a string with a particular value.

dbudworth:
On a total side note, if you wanted to use twirp with a go module based project (that doesn't live in GOPATH), adding:

`replace github.com/twitchtv/twirp => github.com/peak6/twirp v0.0.0-20180919094615-eddc651701575ab8a9ed24c61a995a5bd49d2ddb+incompatible`

to go.mod makes your project use the fork.

I *DO NOT* suggest folks actually build against my fork as I'm not intending to maintain the fork past this pull request.

I'm just posting it merely to unstick anyone who's currently stuck, or in case folks are trying to figure out in general how to override deps.

protoc-gen-twirp/generator.go
dbudworth: Done, makes sense.  Not sure what I was thinking on that initial commit.

dbudworth:
Just checking in, was the bool change what you were expecting?

spenczar:
@dbudworth this looks great, I was just away from github for a few days. Thanks!

Issue #130, opened 2018-09-29
Python 3 client problem


smulube:
This issue relates to https://github.com/twitchtv/twirp/pull/97 I believe.

I've been attempting to generate a python client to interact with a Twirp service written in Go and the generated client is unable to make requests to the server when running under Python 3 (tested with 3.6.6).

As a result of #97, the generated client now includes a line inside the constructor like:

```python
self.__target = server_address.encode('ascii')
```

Now under Python 2 this encode function returns another (encoded) string, however under Python 3, this function returns a distinct `bytes` type. This then breaks the generated `__make_request` function where we do something like this:

```python
req = Request(
    url=self.__target + '/twirp' + full_method,
    data=body,
    ...
```

On the line where we try to concatenate `self.__target` which is now a bytes object with the `'/twirp'` and full_method string, we get: `TypeError: can't concat str to bytes`.

For my purposes which is a single client that I can manually patch and I know is running Python 3, I can do something like this:

```python
req = Request(
    url=self.target.decode('ascii') + '/twirp' + full_method
    data=body,
    ...
```

or I suppose convert the other strings to byte types so they all append together nicely, or even revert the line that added that `encode('ascii')` call when initializing the client, but this feels like a bug to me.

I'm neither a Python or any kind of unicode expert so I don't what the _correct_ solution would be in terms of making this work properly under both Python 2 and 3, however I'd be happy to submit a PR if anyone could offer any guidance.

Tagging @daroot as the author of the #97.

daroot:
Definitely a related problem.  I'm not a unicode expert either, and haven't written a ton of Python 3 that cares about unicode string vs byte strings and whether the httplib is safe to use with one or the other.  

I'm not 100% sure there's a simple solution that reliably works for both python2 and python3 that also guarantees passing something into their respective httplibs in a way that doesn't infect the resulting request and break the binary body.  

It may be necessary to actually break down and write some code that cares about python versions via sys.version or has some other logic for passing either a unicode or bytes type object to Request depending on what that version wants.

smulube:
Have been playing with this in the python shell, I think in the generated client constructor we could either do something like:

```python
if sys.version_info[0] > 2:
    self.__target = server_address
else:
    self.__target = server_address.encode('ascii')
```

which is an attempt to follow the [porting guidelines](https://docs.python.org/3/howto/pyporting.html) which state:

[...]

PR #131, opened 2018-10-01
Check for python version to avoid converting target to bytes in python3

smulube:
*Issue #, if available:*
#130 

*Description of changes:*
This PR attempts to add a little guard around the change added in #97 which safely encodes server URLs in Python 2, which unfortunately causes Python 3 clients to be broken as in Python 3 calling `encode('ascii')` returns a `bytes` type rather than a string which we then cannot concatenate with the twirp prefix string and the specific path when trying to make a request.


spenczar:
Thanks for debugging and addressing this. The changes looks right to me, but I'm not expert enough to really evaluate; I've been burned too many times by Python's string encodings to leap forward on this. I'm pinging a few friends who know more about Python to help advise, hang tight.

smulube:
Sure, sounds great @spenczar. I'm fumbling around with this encoding stuff in python too so would definitely appreciate some more authoritative feedback

PR #133, opened 2018-10-22
Generate Go stubs 

lwc:
Generates a simple stub that provides a type safe and simple way to fake out a server.

### DIFF
--- a/clientcompat/internal/clientcompat/clientcompat.twirp.go
+++ b/clientcompat/internal/clientcompat/clientcompat.twirp.go
@@ -144,6 +144,23 @@ func (c *compatServiceJSONClient) NoopMethod(ctx context.Context, in *Empty) (*E
 	return out, nil
 }
 
+// ==================
+// CompatService Stub
+// ==================
+
+type CompatServiceStub struct {
+	OnMethod     func(context.Context, *Req) (*Resp, error)
+	OnNoopMethod func(context.Context, *Empty) (*Empty, error)
+}
+
+func (s *CompatServiceStub) Method(ctx context.Context, in *Req) (*Resp, error) {
+	return s.OnMethod(ctx, in)
+}
+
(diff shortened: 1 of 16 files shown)


spenczar:
Hi, thanks for the contribution.

Can you provide some background on why you would want this added to Twirp? The stub already exists today through the generated interface; these structs appear (to me) to only add a risk of panics on nil function calls.

spenczar:
I see, thanks for the explanation.

I can understand the desire, but I don't think this is something that should be added directly to Twirp's code generator. We've been guided pretty well by a general principle of minimalism - unless it _must_ be in Twirp, we don't include it.

When I use generated mocks, I usually lean on [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) - a similar tool to `moq`, but it also generates some helpful methods for asserting that methods were called.

My point isn't to get into a debate about whether `counterfeiter` is better than `moq` - by point is Twirp shouldn't really be prescribing this for its users. We should let them pick how they want these mocks to work.

Mock generation using tools like `moq` and `counterfeiter` is easy enough without including it in Twirp that I'd rather leave it up to users. The argument that this avoids adding a dependency is kind of valid, but the dependency isn't required, it's opted-into by users who _want_ the dependency. I don't think there is a problem in chaining off of code generation.

Thanks for the contribution and the explanation!

lwc:
Fair enough - out of interest how do you point counterfeiter to the twirp interfaces? it looks to operate in a similar fashion to moq in that it either requires invocation with the interface name, or a `//go:generate` stanza.

🤔 it might be nice to provide some way to inject a user supplied `//go:generate` stanza during interface generation

lwc:
We'd prefer not to hand write that piece I think - I reckon I might pare this down to a separate protoc plugin that simply generates twirp mocks for a given proto file.

Issue #103, opened 2018-04-26
Use EmitDefaults: true for jsonpb.Marshaler in generated server code.


larrymyers:
protoc-gen-twirp/generator.go#L1012

In generateServerJSONMethod the jsonpb.Marshaler instance should likely include `EmitDefaults: true` as an option.  While Go has a concept of zero values, javascript does not (i.e. the number `0` and `undefined` are very different things).

Let's say a proto Message struct in Go is:

```
type Hat struct {
  Name    string
  Size       int32
  IsFancy bool
}
```

Serializing this Hat instance with the twirp jsonpb.Marshaler does the following:

```
hat := Hat{Name: "", Size: 0}
```

JSON:

```
{}
```

Thoughts?

larrymyers:
@spenczar If you want to slot this into the v6 release I'll happily provide a pull request with the code fix and documentation.

ijhuang:
any chance this will make it into v6?

marioizquierdo:
Some context and comments on this issue.

### Zero and Undefined is not the same

True. In many languages (specially in JavaScript) `0` and `undefined` are not the same thing, but that should not matter as long as you use an autogenerated client. For example, if a field type is a integer, the returned value from the client will always be an integer, never undefined. The network protocol is the one that doesn't differentiate between zero-values and undefined, with the goal of using as little bandwidth as possible. Autogenerated clients must de-serialize the code using the official json-pb mapping protocol: https://developers.google.com/protocol-buffers/docs/proto3#json.

If your service actually needs to differentiate between zero and undefined values, you should add a second boolean property in the Proto file (true if the value exists), or wrap the value with one of the [well known message types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf) (message values can be undefined).

### Manually handling/debugging JSON requests

If you are manually parsing the JSON returned by a Twirp service, you probably find annoying having to handle all those undefined values. If you are making curl requests, you probably find annoying having to guess if the missing fields are zero-values or were not implemented yet. I hear you, this may be a real issue.

Being able to access the services with JSON is just *too convenient*, specially from JavaScript, where there's still not a clear easy-to-use alternative. 

### Twirp works with JSON, but is Protobuf at heart

[...]

PR #134, opened 2018-10-23
Add emit_json_defaults option

tobiaskohlbau:
The package jsonpb which is used to un/marshal json values
into protobuf structures has the option to emit default values
within generated json structures. This CL adds support for
this feature by adding emit_json_defaults as command line flag.
Use for e.g. with `--twirp_out=emit_json_defaults=ENABLE:.`.
Notice that once a value is set this feature is turned on. To
disable it just omit `emit_json_defaults` like for e.g.
`--twirp_out=.`.

### DIFF
--- a/protoc-gen-twirp/command_line.go
+++ b/protoc-gen-twirp/command_line.go
@@ -19,9 +19,10 @@ import (
 )
 
 type commandLineParams struct {
-	importPrefix string            // String to prefix to imported package file names.
-	importMap    map[string]string // Mapping from .proto file name to import path.
-	paths        string            // Paths value, used to control file output directory
+	importPrefix     string            // String to prefix to imported package file names.
+	importMap        map[string]string // Mapping from .proto file name to import path.
+	paths            string            // Paths value, used to control file output directory
+	emitJSONDefaults bool              // Whether to render json fields with zero values.
 }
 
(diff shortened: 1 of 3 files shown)


larrymyers:
Looks like this would provide a way forward for my request in #103.

lvht:
👍 I need these feature as well. However, as the twirp does not use go mod, I think @tobiaskohlbau should remove the `go.mod` and `go.sum` to make this PR accepted.

lvht:
any maintainer is here?

pkieltyka:
pretty key feature. 0 is a real value. thank you for adding this to the generator

tobiaskohlbau:
@lvht thanks for pointing out the accidental added go modules files, I've removed them and rebased the commit.

Hope some maintainer has a point in the future if this could be merged or not.

marioizquierdo:
Unfortunately, this will have to stay in a branch for now. But there may be a possibility of adding this feature to a major new release (e.g. version 6 or 7). This may potentially break things, but it may be useful.

Lets keep the discussion in the issue https://github.com/twitchtv/twirp/issues/103. I added a long comment exploring different points of view: https://github.com/twitchtv/twirp/issues/103#issuecomment-451582978

dpolansky:
Closing this for now. See the discussion in #103 for more context.

PR #143, opened 2019-01-24
Make twirptest a public package

### LINKED ISSUE #142: Make twirptest a public package instead of internal
Given that we are using using `twirptest` in various tests and whatnot (like here: https://github.com/twitchtv/twirp/blob/master/hooks/statsd/statsd_test.go#L23), I think it'd be beneficial to have the package public for individuals who want to write hooks for Twirp.

iheanyi:
Fixes #142 .

This pull request converts `twirptest` into a public package rather than internal so that other Go packages can use it when writing their tests.


iheanyi:
You can check out an example of using this `twirptest` package in a custom hook package here: https://github.com/iheanyi/twirp-opentracing

dpolansky:
Closing in favor of the alternative solution proposed in #142.

Issue #138, opened 2018-12-17
Inconsistent golang type name between protoc-gen-twirp and protoc-gen-go


tomoemon:
There is a golang type name inconsistency between a protoc-gen-twirp plugin and a protoc-gen-go plugin.

The protoc-gen-go plugin generates a struct name with camel case (`item_list` message becomes `itemList` struct), but protoc-gen-twirp plugin generates an interface which has parameters with original message name. (refers a `item_list` struct)

# Steps to reproduce

Create a proto file like below. (`sample.proto`)
```proto
syntax = "proto3";
package sample;

service SampleService {
    rpc Get_v1 (SampleReq_v1) returns (SampleRes_v1);
}
message SampleReq_v1 {}
message SampleRes_v1 {}
```

Generate golang files.
```shell
protoc --go_out=./ --twirp_out=./ ./sample.proto
```

Then, protoc-gen-go plugin generates a struct like this,
```go
type SampleReqV1 struct {
...
}
```
but protoc-gen-twirp plugin generates an interface like this.
```go
type SampleService interface {
	GetV1(context.Context, *SampleReq_v1) (*SampleRes_v1, error)
}
```
As there isn't SampleReq_v1 struct anywhere, failed to build.

# Workaround

Change this line
https://github.com/twitchtv/twirp/blob/master/protoc-gen-twirp/generator.go#L1231
to
```go
name += stringutils.CamelCase(def.Descriptor.GetName())
```

jtvjorge:
Twitch are investigating, thank you for the report!

dpolansky:
Hi @tomoemon,

Sorry for the late turnaround on this. This was fixed in #146 and released in v5.5.2.

Thanks for the bug report.

PR #146, opened 2019-02-14
protoc-gen-twirp: use camel case for Go type name of proto msg

dpolansky:
*Issue #, if available:*
#138
*Description of changes:*
This change updates the protoc-gen-twirp generator to convert the protobuf message names to camel case when it builds the name of that message's Go type. protoc-gen-go handles this edge case properly by emitting the Go type names as camel case.

I added a regression test to `internal/twirptest/snake_case_names`, which fails to compile without the change to the generator.


dpolansky:
@rhysh noticed that the same issue would occur for nested messages. I expanded the test to include nested messages and fixed that issue as well.

Issue #84, opened 2018-02-16
Proposal: Stop using 'OrigName' option when doing jsonpb marshaling


larrymyers:
Referencing this part of the generated code:

https://github.com/twitchtv/twirp/blob/c0c435531e9a1787cec6159bd6262bcf5aa48bea/protoc-gen-twirp/generator.go#L640

Right now the jsonpb.Marshaler is configured to use the original name of the proto Message, instead of doing the default camel casing.

Is the design of twirp that the proto and json serialization should always be equivalent? It would be less work to do idiomatic javascript twirp clients if twirp allowed jsonpb to do its default behavior of lowerCamelCasing Message fields:

https://developers.google.com/protocol-buffers/docs/proto3#json

If this isn't an option for the v5 release, can it be debated for the current v6 work since it would be a breaking change?

spenczar:
Of course, I'm always open to design debates.

The divergence from the spec here is a little odd, I agree, but let's talk a bit about why Twirp accepts JSON payloads: it provides a way to get stuff done when Protobuf payloads aren't available for your client. That happens in two main cases:
 - There's no Twirp client generator for the language, so the client must be hand-written, and hand-writing a Protobuf client is very hard.
 - You're writing a request on the command line with cURL to debug something.

So, JSON shows up when people are hand-writing requests, pretty much. 

Twirp uses `OrigName` because, in practice, people who hand-write clients usually aren't deeply familiar with the JSON-encoded-Protobuf spec. Without `OrigName`, they almost always get serialization - they seemed to be universally surprised by the camelCase remapping. It's much easier to describe how you write the cURL request when you can just say "use the same field names as the proto document."

If we change this, I'm concerned that we will make life harder for people hand-writing JSON clients, which is bad because those are the primary user of JSON payloads.

The tradeoff is that we would probably make things _easier_ for people who are writing client generators, and who are working on JSON clients. I don't weight that benefit terribly heavily - if you're making a client generator, you should just go the full distance and generate protobuf clients. I think client generators don't really even need to make working JSON clients at all; the Go one does to serve as reference, but the Python one just speaks Protobuf, for example.

As you pointed out, this would be a breaking change - and it would be a very significant one. v6 servers would be unable to deserialize JSON requests at all from v5 clients, which is pretty brutal. We would need _very_ strong reasons for doing this to overcome that negative.

larrymyers:
[...]

PR #151, opened 2019-02-28
Correct docs relating to JSON serialization.

itspage:
Names are actually `snake_case`.

As per issue: https://github.com/twitchtv/twirp/issues/84

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -39,11 +39,12 @@ for more details.
 
 ### JSON serialization
 
-The JSON format should match the
+The JSON format generally matches the
 [official spec](https://developers.google.com/protocol-buffers/docs/proto3#json)'s
-rules for JSON serialization. In a nutshell: names are `camelCased`, _all_
+rules for JSON serialization.  In a nutshell: _all_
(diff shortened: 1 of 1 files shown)


spenczar:
Names aren't `snake_case`; they just match exactly the proto name, which most people write in snake case. 

I'm not sure I agree that this violates the jsonpb spec. It's a clearly-supported option in the Go jsonpb marshaler, and is mentioned in the spec as a valid encoding ("Parsers accept both the lowerCamelCase name (or the one specified by the json_name option) and the original proto field name.").

It's hard to be exact here because the JSON encoding spec is just not very precise, for sure, but where do you see a disagreement with that spec?

itspage:
@spenczar you are correct, names aren't `snake_case`, they do match the proto name.
I've updated the docs accordingly.

The disagreement with the spec is here: https://developers.google.com/protocol-buffers/docs/proto3#json

> Message field names are mapped to lowerCamelCase and become JSON object keys.

marioizquierdo (approved):
Thanks for the clarification, and thanks @spenczar for the comments 👍

PR #153, opened 2019-03-07
Update build

peter-edge:
This PR does a few things:

- Uses Golang Modules to install `errcheck, protoc-gen-go, and protoc-gen-gofast`. This is very simple as of Go 1.11 and allows pinning of versions.
- Install `protoc` from GitHub Releases, instead of relying on a system-wide installation.
- Install `dep` from GitHub Releases, instead of relying on a system-wide installation.
- Upgrades golang/protobuf and gogo/protobuf to the latest 1.3.0 and 1.2.1 respectively.
- Upgrades stretchr/testify to 1.3.0.
- Add a new `vendor_update` target that calls `dep ensure -update` using the installed `dep`.
- `make vendor_update`.
- Remove `_tools`. I know `retool` is a Twitch project, and this might not feel great, but with Golang Modules (available for two Golang versions now, so this feels OK) it's not needed as much anymore, and this simplifies the build a lot. Note that Twirp itself remains compatible with previous Golang versions, this is just a build concern.
- Remove the `setup` target as it is no longer needed.
- `make generate` to pick up new golang/protobuf and gogo/protobuf changes.
- Removes 1.9 and 1.10 from `.travis.yml` since they will not work with the build anymore.
- Updates the documentation to say Go 1.11+.
- Sets `go_import_path` in `.travis.yml` to make builds work with forks.
- Have Travis call `make` so that CI matches development locally.

With these changes, the build is entirely self-contained now, with no need for system tools other than git, go, curl, make. You could go a step further and install Golang itself as part of the build on a specific version if you wanted to using the same technique in the Makefile as done below.

Note that in the Makefile, it is possible to call `errcheck, protoc`, etc because the `PATH` is set in the Makefile. Even if there is a system-wide installation, these take precedence, so the system-wide installation will not be called. This removes the need to do i.e. `$(ERRCHECK)` and you can instead call `errcheck` (or more to the point, `protoc` in your scripts) with no issue.

The `.tmp` directory is split up by OS/Arch to make this easy to use with Docker.

This is just a suggestion - feel free to reject this diff obviously. I would recommend checking it out and playing with it first if you want to consider merging it.

```
# not that you would need instructions but just for copy/paste purposes
cd $GOPATH/src/github.com/twitchtv
mv twirp twirp.bk
git clone https://github.com/peter-edge/twirp
cd twirp
git checkout update-build
make
```


peter-edge:
I know the diff is huge due to updating `vendor` and deleting `_tools`, these are the two files you'll care most about:

https://github.com/peter-edge/twirp/blob/update-build/Makefile
https://github.com/peter-edge/twirp/blob/update-build/.travis.yml

Makefile
```diff
@@ -1,21 +1,91 @@
-RETOOL=$(CURDIR)/_tools/bin/retool
-PATH := ${PWD}/bin:${PWD}/ENV/bin:${PATH}
+SHELL := /bin/bash -o pipefail
+UNAME_OS := $(shell uname -s)
+UNAME_ARCH := $(shell uname -m)
+
+TMP_BASE := .tmp
```
GarethLewin: For clarity, I would name this `.tools` and all the variables `.tmp` because (unless I am reading this wrong) the folder isn't really temporary and lasts as long as you work on twirp.

Makefile
```diff
@@ -1,21 +1,91 @@
...
+TMP := $(TMP_BASE)/$(UNAME_OS)/$(UNAME_ARCH)
+TMP_BIN = $(TMP)/bin
+
+DEP_VERSION := 0.5.0
```
GarethLewin: Minor complaint, this means that if we move to a newer DEP version cruft will start to build up with multiple versions of DEP in the folder. This is true for all the apps.

Makefile
```diff
@@ -1,21 +1,91 @@
...
+endif
+ifeq ($(UNAME_ARCH),x86_64)
+DEP_ARCH := amd64
+endif
+$(DEP):
+	@mkdir -p "$(TMP_BIN)"
+	curl -sSL "https://github.com/golang/dep/releases/download/v$(DEP_VERSION)/dep-$(DEP_OS)-$(DEP_ARCH)" -o "$(DEP)"
+	@chmod +x "$(DEP)"
+
+PROTOC_VERSION := 3.7.0
```
GarethLewin: For clarity I would move all the `VERSION` constants to the top. I realize this is bike shedding but I think it would make it easier to understand what this makefile installs and at what version than having to search throughout the file.

Makefile
```diff
@@ -1,21 +1,91 @@
...
+	@rm -rf $(PROTOC_GEN_GO_TMP)
+
+PROTOC_GEN_GOFAST_VERSION := v1.2.1
+PROTOC_GEN_GOFAST := $(TMP_BIN)/protoc-gen-gofast
+$(PROTOC_GEN_GOFAST):
+	$(eval PROTOC_GEN_GOFAST_TMP := $(shell mktemp -d))
+	@cd $(PROTOC_GEN_GOFAST_TMP); echo module tmp > go.mod; GO111MODULE=on go get github.com/gogo/protobuf/protoc-gen-gofast@$(PROTOC_GEN_GOFAST_VERSION)
+	@rm -rf $(PROTOC_GEN_GOFAST_TMP)
+
+export GOBIN := $(abspath $(TMP_BIN))
```
GarethLewin: This aliased GOBIN seems unneeded, `export PATH := $(abspath $(TMP_BIN)):$(PATH)` seems just as clear. But again minor.

peter-edge: It's needed so that the `go get` and `go install` calls install to the correct location, it makes it so that `errcheck, protoc-gen-go, protoc-gen-gofast, protoc-gen-twirp, protoc-gen-twirp_python` all install to this directory.

Makefile
peter-edge: It's mean to be temporary, and cleaned easily - I added `make clean` that just `rm -rf .tmp` which is a common thing I do. It's .gitignored, so this works nicely.

Makefile
peter-edge: They won't - I purposefully write them to the same location `$TMP_BIN/dep`. You can put the version as part of the binary name, which I have done in the past, but that means you can't call `dep`, instead you have to call `$(DEP)`, and for example with `protoc`, this is problematic with how the scripts are set up (the scripts invoke `protoc`, not `protoc-3.7.0`, since they don't have access to this version information).

If you upgrade the tools, it is good to call `make clean`.

Makefile
peter-edge: Happy to do it, I would disagree personally because this groups things nicely - if I care about `protoc`, everything is here. But can do this, no problem.

peter-edge:
I updated the build assets to go to `.tmp` as well - this makes it easy when operating in multiple environments, ie Mac and a Docker container. I addressed the comment as well.

Makefile
peter-edge: Done

Makefile
GarethLewin: I missed that, but I think that this is adding unknown tribal knowledge complexity to someone working on the project. I would advocate at least a section in the readme about this. Essentially this is rolling out a mini package manager without all the hard parts that make writing a package manager hard.

For example if I work on twirp and pull a new copy of master I now need to verify myself if the makefile changed a version, and if it did I have to clean `.tmp` myself.

peter-edge: Just my opinion: having just gone through building Twirp the first time, this system should already be **way** easier than the system using retool - I had to fight a bunch of things to get deps updated. You're right in that it does it's own package management, and the simplicity of it is the killer feature - nothing to learn, it's basically hermetic (outside of git, go, curl, make), and it works easily on both Darwin and Linux.

I can add something to CONTRIBUTING.md if you think it's important.

peter-edge: One other note: This is no different than the current setup in this manner. When I check out a new copy of master right now, I need to run `make setup` to verify that `retool build` is called (or I need to know to call `retool`). `setup` isn't called before every relevant target.

peter-edge: I've now fixed the problem - each tool now depends on a file `$(TMP)/versions/NAME/VERSION` instead of the actual binary. When installing, the `NAME` directory is deleted, and the file is created with `touch`. This makes it so that if versions switch, you will be updated, including if you switch back and forth, but the binary will still have the same name.

I've verified this works locally as expected by switching versions back and forth, and making sure that the install happens when a version is changed, but not at other times.

peter-edge:
I also removed the `compat` file targets - these are currently not specified in a way that would result in a rebuild if any of them changed. Since the rebuild of each is so quick, you might as well make it easy and just rebuild them before each test. You could have more complicated Makefile rules to make sure they only get rebuilt if source files change, but you get into complicated territory to save 1-2 seconds, and rebuilding is likely more reliable.

.gitignore
```diff
@@ -1,9 +1,7 @@
 *.test
-/bin
+/.tmp/
 
 clientcompat/pycompat/ENV
 *.pyc
 
-build
```
spenczar: This is used in generating the documentation website, and should not be removed.

Makefile
```diff
@@ -1,51 +1,131 @@
...
-generate:
-	PATH=$(CURDIR)/_tools/bin:$(PATH) GOBIN="${PWD}/bin" go install -v ./protoc-gen-...
-	$(RETOOL) do go generate ./...
+$(TMP_BIN):
+	@mkdir -p $(TMP_BIN)
 
-test_all: setup test_core test_clients
+$(TMP_VENV):
+	@mkdir -p $(dir $(TMP_VENV))
+	virtualenv $(TMP_VENV)
```
spenczar: Virtualenv needs to be installed, or called out as a dependency (probably the latter).

Makefile
spenczar: > One other note: This is no different than the current setup in this manner. When I check out a new copy of master right now, I need to run make setup to verify that retool build is called (or I need to know to call retool). setup isn't called before every relevant target.

To be clear, `setup` doesn't need to be called every time. Retool checks whether tools are synced to the right version on [every `retool do` invocation](https://github.com/twitchtv/retool/blob/master/main.go#L52-L54), so they're always at the right version. But yes, the touched filenames bring this in to your Makefile version, too.

spenczar:
I'm closing this, but the ideas are still useful - thanks a lot, Peter.

peter-edge:
Note this does use a local install of the WKT. They’re in a temp include
directory. If for some reason protoc is still going back to system-wide
install, then just include them.

All would be handled with prototool.

spenczar:
You are right, of course. I misread the line `@rm -rf $(TMP)/include/google/protobuf`; it occurs _before_ the archive is unzipped, not after.

PR #160, opened 2019-03-15
Get panic cause as error in Error hooks

marioizquierdo:
Currently, service method handlers make sure that panics still call Error hooks. However, the `twirp.Error` passed to the error hook has no information about the panic, it is simply an `Internal` error with Msg: "Internal service panic".

This PR adds a `Cause` to the error passed to the Error hooks, so it can be unwrapped with `pkg/errors.Cause`. The resulting error response is the same, with Msg: "Internal service panic", the information about the panic is only available in the server through accessing the cause.

```go
// Given hooks with Error handling ...
hooks := &twirp.ServerHooks{
	Error: func(ctx context.Context, twerr twirp.Error) context.Context {		
		// Some errors are wrapped as Internal errors. Panics are also wrapped here.
		err := pkgerrors.Cause(twerr)

		// Before this PR
		twerr.Msg() //=> "Internal service panic"
		err.Error() //=> "twirp error internal: Internal service panic"

		// After this PR
		twerr.Msg() //=> "Internal service panic"
		err.Error() //=> shows the original error, or "panic: %v" if the panic value is not an error
	},
}
```

In addition, I made a little bit of refactor to simplify and explain the code that handles panics.


_tools/src/golang.org/x/tools/go/ast/astutil/imports.go
```diff
@@ -14,26 +14,26 @@ import (
 )
 
 // AddImport adds the import path to the file f, if absent.
-func AddImport(fset *token.FileSet, f *ast.File, ipath string) (added bool) {
-	return AddNamedImport(fset, f, "", ipath)
+func AddImport(fset *token.FileSet, f *ast.File, path string) (added bool) {
+	return AddNamedImport(fset, f, "", path)
```
marioizquierdo: Sorry, all of this appeared when I first ran `make test_core`.
I wonder if I can `git reset` this, or if we should keep this diff around to keep tools up to date ...

marioizquierdo:
Fixes #163

internal/twirptest/service.twirp.go
```diff
@@ -565,33 +540,68 @@ func twirpErrorFromIntermediary(status int, msg string, bodyOrLocation string) t
...
+
+func (e *wrappedError) Cause() error  { return e.cause }
+func (e *wrappedError) Error() string { return e.prefix + ": " + e.cause.Error() }
+
+// handlePanics makes sure that panics during rpc methods respond with Twirp Interna
+// error responses (status 500) and trigger hooks with the panic wrapped as an error.
+// The panic is not intercepted and should be handled in middleware to avoid empty responses.
+func handlePanics(ctx context.Context, resp http.ResponseWriter, hooks *twirp.ServerHooks) {
+	if r := recover(); r != nil {
+		err := errFromPanic(r)                                                 // allow to inspect error on Error hooks
```
spenczar: These comments are overflowing way off to the right side. Could we get them to stay under ~100 characters, either by moving them above the lines they describe or by not vertically aligning them?

internal/twirptest/service_test.go
```diff
@@ -802,7 +803,7 @@ func TestPanickyApplication(t *testing.T) {
 	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if r := recover(); r == nil {
-				t.Error("http server never panicked")
+				t.Error("http server never panicked") // also make sure it did panic
```
spenczar: I don't think this comment is useful.

internal/twirptest/service_test.go
```diff
@@ -834,6 +835,60 @@ func TestPanickyApplication(t *testing.T) {
...
 
+// If an application panics in its handler, the Error hooks should trigger and have access to the panic source.
+func TestPanicsTriggerErrorHooks(t *testing.T) {
+	errHookCalled := false
+	errHook := &twirp.ServerHooks{
+		Error: func(ctx context.Context, twerr twirp.Error) context.Context {
+			errHookCalled = true
+			// The error should have a .Cause containing the panic value for inspection
+			err := pkgerrors.Cause(twerr)
+			if err.Error() != "This Is Fine Meme" {
```
spenczar: Rather than testing the error string, you should be able to test whether `err == panicValue` directly.

internal/twirptest/service_test.go
```diff
@@ -834,6 +835,60 @@ func TestPanickyApplication(t *testing.T) {
...
+				t.Error("http server never panicked") // also make sure it did panic
+			}
+		}()
+		s.ServeHTTP(w, r)
+	})
+
+	server := httptest.NewServer(h)
+	defer server.Close()
+
+	client := NewHaberdasherJSONClient(server.URL, http.DefaultClient)
```
spenczar: Would be good to test both JSON and Protobuf code paths.

internal/twirptest/service.twirp.go
```diff
@@ -381,32 +365,23 @@ func writeError(ctx context.Context, resp http.ResponseWriter, err error, hooks
...
-		// Silently ignoring the error is our least-bad option. It's highly
-		// likely that the connection is broken and the original 'err' says
-		// so anyway.
+		// Calling the Error hook is unacceptable because it would trigger twice.
+		// The best option is to ignore writeErr, the original err is being handled anyways.
 		_ = writeErr
 	}
 
-	callResponseSent(ctx, hooks)
+	callResponseSent(ctx, hooks) // ResponseSent hook
```
spenczar: Remove this comment

internal/twirptest/service.twirp.go
```diff
@@ -381,32 +365,23 @@ func writeError(ctx context.Context, resp http.ResponseWriter, err error, hooks
...
+	resp.WriteHeader(statusCode)                          // set HTTP status code and send response
 
 	respBody := marshalErrorToJSON(twerr)
 	_, writeErr := resp.Write(respBody)
 	if writeErr != nil {
-		// We have three options here. We could log the error, call the Error
-		// hook, or just silently ignore the error.
-		//
-		// Logging is unacceptable because we don't have a user-controlled
+		// Most likely the connection is broken.
```
spenczar: I prefer the old comment that was here, could you revert this change?

internal/twirptest/service.twirp.go
```diff
@@ -381,32 +365,23 @@ func writeError(ctx context.Context, resp http.ResponseWriter, err error, hooks
 
 	statusCode := twirp.ServerHTTPStatusFromErrorCode(twerr.Code())
 	ctx = ctxsetters.WithStatusCode(ctx, statusCode)
-	ctx = callError(ctx, hooks, twerr)
+	ctx = callError(ctx, hooks, twerr) // Error hook
```
spenczar: Remove this comment

internal/twirptest/service.twirp.go
```diff
@@ -565,33 +540,68 @@ func twirpErrorFromIntermediary(status int, msg string, bodyOrLocation string) t
...
-	msg   string
-	cause error
+	prefix string
+	cause  error
+}
+
+func (e *wrappedError) Cause() error  { return e.cause }
+func (e *wrappedError) Error() string { return e.prefix + ": " + e.cause.Error() }
+
+// handlePanics makes sure that panics during rpc methods respond with Twirp Interna
```
spenczar: typo: `Internal`

internal/twirptest/service.twirp.go
```diff
@@ -565,33 +540,68 @@ func twirpErrorFromIntermediary(status int, msg string, bodyOrLocation string) t
...
+// error responses (status 500) and trigger hooks with the panic wrapped as an error.
+// The panic is not intercepted and should be handled in middleware to avoid empty responses.
```
spenczar: I'd phrase this differently. The panic is definitely intercepted, we just re-raise it.

Empty responses are kind of a separate issue that we should be able to deal with by flushing correctly.

internal/twirptest/service.twirp.go
marioizquierdo: They seem to auto-align when generating code. I didn't realize that was happening. Fixing ...

internal/twirptest/service.twirp.go
marioizquierdo: Would it make sense to rename `callError` to `callErrorHook`? and `callResponseSent` to `callResponseSentHook`? etc.  No big deal, I just found "callError" to be a little misleading.

internal/twirptest/service.twirp.go
marioizquierdo: Yes. I'll phrase it differently to be more clear. I found the empty response issue while working on this, it was not as clear to me before.

marioizquierdo: For reference, the empty responses issue is tracked here https://github.com/twitchtv/twirp/issues/163

internal/twirptest/service_test.go
```diff
@@ -834,6 +835,64 @@ func TestPanickyApplication(t *testing.T) {
...
+	clients := map[string]Haberdasher{
+		"protobuf": NewHaberdasherProtobufClient(server.URL, http.DefaultClient),
+		"json":     NewHaberdasherJSONClient(server.URL, http.DefaultClient),
+	}
+	for name, c := range clients {
+		_, err := c.MakeHat(context.Background(), &Size{Inches: 1})
+		if err == nil {
+			t.Fatalf("%s client: err was expected for panicking handler, found nil", name)
+		}
+		if !errHookCalled {
```
spenczar: Need to reset this variable to `false` after the first loop iteration

_tools/src/golang.org/x/tools/go/ast/astutil/imports.go
spenczar: I think we need to rethink our build tooling in general, maybe moving to doing everything in Docker. I'll just accept these changes for now.

internal/twirptest/service_test.go
spenczar: (fixed it myself)

Issue #145, opened 2019-02-14
Proposal: Client-Side Hooks in Go


iheanyi:
Similar to how the server-side hooks area  thing, it'd be beneficial to have client-side hooks as well for automating things such as client-side logging, retries, etc.

Here's a rough draft of the API that I've thought up.

```go
type ClientHooks struct {
  // RequestPrepared is called as soon as a request has been created and before it has been sent 
 // to the Twirp server.
  RequestPrepared func(context.Context, *http.Request) (context.Context, error)

 // RequestFinished (alternatively could be named ResponseReceived and take in some response from 
 // the server as well) is called after a request has finished sending. Since this is terminal, the context is 
 // not returned. 
 RequestFinished func(context.Context)

 // Error hook is called whenever an error occurs during the sending of a request. The Error is passed
 // as an argument to the hook.
  Error func(context.Context, twirp.Error) context.Context 
}
```

Looking forward to hearing y'alls thoughts!

dpolansky:
Hi @iheanyi,

Sorry for the late turnaround on this issue. Twirp has server hooks because 1.) you can run into issues defining HTTP server middleware by wrapping `http.ResponseWriter` (more context in https://github.com/golang/go/issues/18997) and 2.) the provided server hooks provide functionality that would be difficult to emulate with HTTP server middleware.

With this in mind, we should consider how client hooks would differ from HTTP client middleware (using a type that fulfills the Twirp `HTTPClient` interface) and whether there is enough of a use case to warrant expanding the public API.

For the `RequestPrepared`, `RequestFinished`, and `Error` hooks that you propose, I think they can all be implemented as a custom `HTTPClient`. Are there other use cases that you are interested in where HTTP client middleware is not an appropriate solution?

iheanyi:
Hey there @dpolansky, thanks for circling back on this. I was getting this idea as I was working on the `twirp-opentracing` package. I think you're right in that the `twirp.HTTPClient` interface can be a good middleground. I was confused by this at first because while working on `twirp-opentracing`, I didn't realize that interface existed because it's only defined within the generated code (as seen [here](https://github.com/iheanyi/twirp-opentracing/blob/master/vendor/github.com/iheanyi/twirptest/service.twirp.go#L336-L346)), rather than in the standard library. You can see my implementation of a client middleware [here](https://github.com/iheanyi/twirp-opentracing/blob/master/trace_http_client.go). 

[...]

PR #164, opened 2019-03-25
[DO NOT MERGE] RFC: Client Hooks Proof of Concept

iheanyi:
*Issue #, if available:* #145 

*Description of changes:* This pull request makes a few changes. It modifies `twirptest/service.twirp.go` directly to explore how things could change if we were to introduce client-side hooks in a backwards compatible way. Notice how some type signatures in tests and such changed, but the underlying function calls for generating clients do not change. This is a way that Twirp can add in client-side hooks in an opt-in way and not break existing clients.


spenczar:
Closing this in favor of #198. Thanks for the sketch, I think it helped us get in the right direction!

Issue #158, opened 2019-03-13
Proposal: Add PathPrefix method to generated TwirpServer type


gorzell:
I ran into some challenges working on code to try and improve the experience of wiring up a set of Twirp RPC services via a function, along the lines of:

```go
func Register(mux http.ServeMux, twirpServices ... twirp.TwirpServer) {
	for _, service := range twirpServices {
		mux.Handle(service.Prefix(), service)
	}
}
```

I had based this on the [docs](https://twitchtv.github.io/twirp/docs/mux.html) about adding services to an existing `http.ServeMux`, but ran into the following nuances of how `TwirpServer` and `const *Prefix` are currently implemented:

* When you call `New` on a generated service you get a `TwirpServer`. This is good thing, however the interface definition is **inside** the generated code and its package. This means that you can't refer to the interface in more general context. There doesn't seem to be anything implementation specific about this interface that would lead to it needing to be part of the generated code, so I am wondering if it couldn't be moved up into the `twirp` package.
* Having a `const` with the path prefix created as part of code generation is super useful, but it can be hard to get to. Given that the description of `TwirpServer` is "TwirpServer is the interface generated server structs will support: they're HTTP handlers with additional methods for accessing metadata about the service," I wonder if it wouldn't make sense to add a `Prefix()` method to the interface and then generate an implementation that returns the `const` value. The prefixs would then be accessible in a more generic way.

I think these changes are generally beneficial in addition to enabling the example that I started with. However I am fairly new to this code base, so if there are reasons that it is this way or other things that I am missing, please let me know. I am also happy to create a PR that implements this if it is deemed to be a good path forward.

gorzell:
I think most of the first point is answered in: https://github.com/twitchtv/twirp/issues/96

dpolansky:
Thanks for the proposal. I agree that the comment by @spenczar in #96 does a great job describing the problems Twirp would run into if we moved the `TwirpServer` interface out of the generated code. I agree with your point about the path prefix being hard to get to in a generic way, and adding that to the generated code seems fairly painless. I'm going to shop this by the other maintainers this week and see what they think.

spenczar:
Right, you can define an interface type that covers your needs, and not need to reference the generated one, since interfaces that are identical are mutually assignable.

The second concern, on prefixes, is more interesting to me. I agree that it is painful to get generic access to prefixes today, but would like to understand why this would be useful. Could you give an example?

gorzell:
[...]

PR #167, opened 2019-03-26
Add PathPrefix to the generated TwirpServer interface.

gorzell:
*Issue:* #158

*Description of changes:*
The generated code implements the new function by returning the existing
const for this value. Having an interface method provides consistent
access for all TwirpServer instances, rather than having to use a
instance specific constant.

**Note:** If the `Dockerfile` isn't generally useful, I am happy to revert that commit.


protoc-gen-twirp/generator.go
```diff
@@ -352,6 +352,10 @@ func (t *twirp) generateUtils() {
 	t.P(`	// ProtocGenTwirpVersion is the semantic version string of the version of`)
 	t.P(`	// twirp used to generate this file.`)
 	t.P(`	ProtocGenTwirpVersion() string`)
+	t.P(`// PathPrefix is used for all URL paths on the generated server.`)
```
spenczar: I'm picky about comments. Could you phrase this as follows:

```go
// PathPrefix returns the HTTP URL path prefix for all methods handled by this
// service. This can be used with an HTTP mux to route twirp requests
// alongside non-twirp requests on one HTTP listener.
```

I realize that you've copied the comment that's on the `const`, but I think that comment could be improved too.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.12
```
spenczar: Should we use this in our travis builds, too?

gorzell: If you mean the GoLang version, I think it is always nice to be current.  If you mean the container itself, I don't see why not, but I also don't know that much about how travis works.

spenczar: I meant the container itself.

I think it's usually important that the way we do dev work mirrors exactly the way we do CI.

Could you revert the commit that adds the Docker container? I think it might be a good idea, but I'd like to do it more comprehensively, covering each of the Go versions we want to support, that sort of thing.

protoc-gen-twirp/generator.go
gorzell: I don't mind changing this.  Should I change the comment on the `const` as well?

gorzell:
@spenczar I think I covered everything in: https://github.com/twitchtv/twirp/pull/167#pullrequestreview-219097865, and I reverted the `Dockerfile`.

I think the only open items at this point are:
* Was I too lazy/literal in the way I changed the Mux test?
* Should I change the `const` godoc to match the `PathPrefix()` version?

example/service.twirp.go
```diff
@@ -363,6 +367,10 @@ type TwirpServer interface {
 	// ProtocGenTwirpVersion is the semantic version string of the version of
 	// twirp used to generate this file.
 	ProtocGenTwirpVersion() string
+	// PathPrefix is used for all URL paths on the generated server.
```
spenczar: Please re-run code generation with the new comment.

spenczar:
> Was I too lazy/literal in the way I changed the Mux test?
No, I really like the way you've done it.

> Should I change the const `godoc` to match the `PathPrefix()` version?
I don't think that's necessary in this revision.

gorzell:
@spenczar I think this should be ready to be merged now.

spenczar:
Thanks for the nice design, @gorzell. I've cut a new release, v5.6.0, which adds this feature, and I've updated the documentation website.

Issue #162, opened 2019-03-20
Panics result in empty responses unless they are intercepted in middleware


marioizquierdo:
Auto-generated code for Golang Twirp services seem to be handling panics:
https://github.com/twitchtv/twirp/blob/311d20ef3700822f5683641c9136d85fd5e14422/protoc-gen-twirp/generator.go#L1020

### Expected behavior

The code suggests that panics should result in a `twirp.Internal` error like this:
```go
# curl -X POST http://localhost:9090/twirp/mypkg.HelloWorld/Hello ....
{"code": "internal", "msg": "Internal service panic"}
```

### Actual behavior

However, they result in an empty response. Doing a curl request looks like this:
```sh
# curl -X POST http://localhost:9090/twirp/mypkg.HelloWorld/Hello ....
curl: (52) Empty reply from server
```

### Intercepting the panic in middleware

Adding middleware to intercept the panic allows the response to properly flush:

```go
func InterceptPanics(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		defer func() {
			if p := recover(); p != nil {
				// do whatever, but don't re-panic
			}
		}()
		h.ServeHTTP(w, r)
	})
}
```

### Using w.(http.Flusher).Flush()

Maybe Twirp should make sure that the internal error response should be flushed before re-throwing the panic?

I tried changing the auto-generated code, adding `w.(http.Flusher).Flush()` right before re-throwing the panic and it generates a valid response, however the http connection was in a chunked mode and curl was awaiting for more chunks or 0x00 0x00 sequence. Instead it got a connection closed.

```
# curl -X POST http://localhost:9090/twirp/mypkg.HelloWorld/Hello ....
curl: (18) transfer closed with outstanding read data remaining
{"code": "internal", "msg": "Internal service panic"}
```

spenczar:
I think we should Flush in the generated code.

We should be able to explicitly set `Content-Length`, which avoids switching us into chunked transfer encoding, as documented in [http.ResponseWriter.Write](https://godoc.org/net/http#ResponseWriter).

spenczar:
Fixed in #168, release v5.7.0.

PR #168, opened 2019-04-02
Set Content-Length of panic responses and flush them

spenczar:
*Issue #, if available:*
#162 
*Description of changes:*
This changes panic handling in generated Twirp servers. They now calculate the length of the response, set it in the Content-Length header of the response, and explicitly flush.

When a service panics, we write an error into the `http.ResponseWriter` and re-raise the panic. If nothing up the `http.Handler` stack intercepts that panic, then the write might not get flushed in time, resulting in a confusing "EOF" error message for the client.

We need to flush the error explicitly. But if we only do that, then the Go HTTP library will switch into chunked transfer encoding mode. If that happens, the client doesn't know how many bytes to wait for in the error response, and the panicking server will close the HTTP response uncleanly, without terminating the chunked transfer response. As a result, if we just flush, the client _still_ just gets a
confusing "EOF" message, since it's left unsure of whether the chunked transfer is done.

The right thing is to explicitly set the content length of the response, which prevents the Go HTTP library from switching to chunked-transfer, and which informs the client of exactly how many bytes to expect. Only then can we flush.


protoc-gen-twirp/generator.go
```diff
@@ -591,15 +591,26 @@ func (t *twirp) generateUtils() {
...
-	t.P(`    writeError(ctx, resp, twerr, hooks)`)
+	t.P(`	if r := recover(); r != nil {`)
+	t.P(`		// Wrap the panic as an error so it can be passed to error hooks.`)
+	t.P(`		// The original error is accessible from error hooks, but not visible in the response.`)
+	t.P(`		err := errFromPanic(r)`)
+	t.P(`		twerr := &internalWithCause{msg: "Internal service panic", cause: err}`)
+	t.P(`		// Set Content-Length of response to avoid switching to chunked transfer`)
+	t.P(`		// encoding when we flush, and so the client knows that it has received the`)
+	t.P(`		// complete response when we flush.`)
+	t.P(`		length := len(marshalErrorToJSON(twerr))`)
```
spenczar: We end up calling `marshalErrorToJSON` twice in this code path - once here, and once in `writeError`. I could refactor to avoid that, but wanted the first revision to be clear, so I left in the double marshal.

internal/twirptest/service_test.go
```diff
@@ -1200,3 +1201,40 @@ func TestContextValues(t *testing.T) {
 		t.Errorf("Client err=%q", err)
 	}
 }
+
+func TestPanicFlushing(t *testing.T) {
+	h := PanickyHatmaker("bang!")
+	s := httptest.NewUnstartedServer(NewHaberdasherServer(h, nil))
+	defer s.Close()
+	s.Config.ErrorLog = testLogger(t)
```
marioizquierdo: Is not clear what is being tested or setup here. Maybe this line needs a comment explaining what to expect?

protoc-gen-twirp/generator.go
```diff
@@ -591,15 +591,26 @@ func (t *twirp) generateUtils() {
...
+	t.P(`		resp.Header().Set("Content-Length", strconv.Itoa(length))`)
+	t.P(`		// Actually write the error`)
+	t.P(`		writeError(ctx, resp, twerr, hooks)`)
+	t.P(`		// If possible, flush the error to the wire.`)
+	t.P(`		f, ok := resp.(http.Flusher)`)
+	t.P(`		if ok {`)
+	t.P(`			f.Flush()`)
+	t.P(`		}`)
```
marioizquierdo: style nitpick: mixed spaces and tabs

protoc-gen-twirp/generator.go
marioizquierdo: I think marshaling twice should be fine. Most error responses are small, most panics are unfrequent. If we ever find performance issues with this we can always refactor later.

internal/twirptest/service_test.go
```diff
@@ -1200,3 +1200,31 @@ func TestContextValues(t *testing.T) {
...
+	}
+	twerr, ok := err.(twirp.Error)
+	if !ok {
+		t.Fatalf("expected twirp.Error type error, have %T", err)
+	}
+
+	if twerr.Code() != twirp.Internal {
+		t.Errorf("twirp ErrorCode expected to be %q, but found %q", twirp.Internal, twerr.Code())
+	}
+	if twerr.Msg() != "Internal service panic" {
```
rhysh: Instead of repeating the desired value on the next line, consider `if have, want := twerr.Msg(), "Internal service panic"; have != want {`

internal/twirptest/service_test.go
rhysh: You could change the input to `func(args ...interface{})` and pass in `t.Log` to be more explicit about how `t` can be used.

protoc-gen-twirp/generator.go
rhysh: Why don't we send Content-Length on every error response? This could all move inside `writeError`, which currently calls `resp.WriteHeader` right before calling `marshalErrorToJSON`.

Why don't we send Content-Length on successful responses too? In those cases we definitely have the bytes ready before we write the header (so we can report any marshal errors).

spenczar: That's a good point, @rhysh. I like that. I'll revise to always set Content-Lengths.

spenczar:
This PR now adds Content-Length to _all_ responses. I think this is just better. Thanks for the prod, @rhysh.

Issue #169, opened 2019-04-03
Support Go Modules 


marwan-at-work:
Twirp should be able to support Go Modules for a few reasons: 

1. Pinning dependencies correctly when GO111MODULE will be set to `on` by default in the next release of Go 1.13 

2. It will make the onboarding experience to twirp a little bit easier because you'll be able to get binaries such as protoc-gen-twirp at an exact version without having to use `retool`. For example: `go get github.com/twitchtv/twirp/protoc-gen-twirp@v5.6.0` 

3. There will be no need to vendor dependencies on both the twirp side as well as the user's side. 

I noticed that there was already an issue about adding go.mod/go.sum and the reason for closing it was this comment: https://github.com/twitchtv/twirp/issues/140#issuecomment-452401108

In particular: 

> the transition will involve quite a bit more than just adding go.mod and go.sum files as import paths will need to change

For that reason, I [created a tool](github.com/marwan-at-work/mod) to automatically upgrade import paths. 

I have a draft PR to demonstrate that all import paths are upgraded and that the tests past (at least locally for me). 

The PR is in draft because A. I'm not familiar with Twirp's roadmap and B. I imagine a lot of documentation needs to be updated to show the new flow of getting started. 

Please feel free to keep this open until Twirp officially supports module, and also feel free to add all the TODOs that need to be accomplished before rendering this issue resolved.

spenczar:
I agree that we clearly need to eventually move to Go modules. My biggest concern is the migration path. I think we need to go very carefully here. I'd like to make three design constraints explicit:

First, we need to support the last two minor versions of Go. Today, that's 1.10, 1.11, and 1.12. Go 1.10 doesn't have module support directly, so until Go 1.13, we would need to "fake" modules by putting code in an explicit `github.com/twitchtv/twirp/v6` folder. Perhaps that means it's best to just wait for Go 1.13 to release module support - that's OK with me.

Second, we cannot mandate users (even those in Go 1.12) adopt modules to use Twirp v6 generated code. There are two prongs to this argument. First, adopting modules is not easy: it requires changes to users' build systems, and involves a change in development practices and workflows. It's not a simple switch. Second, if we mandated module usage for generated code to work, that mandate is viral. It affects not just users who run Twirp services, but _all of their clients_. Those two arguments combine to mean that mandatory module adoption would be an onerous burden on too many applications - many of whom (the clients) might not even be using Twirp themselves directly. It's too rude to force them to change their builds to be clients of some other service.

[...]

PR #170, opened 2019-04-03
Add Go Modules

marwan-at-work:
This PR demonstrates support for Go Modules. It doesn't have to be merged as it just shows a working version of twirp with modules. Although it can be merged if the maintainers are comfortable with the change :) 

The versions in the `go.mod` file were generated based on the existent Gopkg.toml file. In particular, the gogo/protobuf package seems quite outdated. 

Adding support for Go Modules for programs that are tagged > 1 are considered backwards incompatible and therefore they require a new version upgrade. Quite unfortunate to introduce a whole new version just because you have module support, but at the same time it ensures that people living in the old world can still function correctly by just using the previous major version. 

I have not deleted the vendor folder nor the *.toml files as they would make this PR look too scary. 

Updates https://github.com/twitchtv/twirp/issues/169

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ go:
 - 1.11.x
 - 1.12.x
 - tip
-environment:
-  - GO111MODULE=on
+env:	
+  global:
+    - GO111MODULE=on
 script:
 - go install ./...
 - go test -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 sudo: false
 language: go
 go:
-- 1.9.x
-- 1.10.x
 - 1.11.x
+- 1.12.x
 - tip
+environment:
+  - GO111MODULE=on
 script:
 - go install ./...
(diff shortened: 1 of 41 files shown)


.travis.yml
```diff
@@ -1,10 +1,12 @@
 sudo: false
 language: go
 go:
-- 1.9.x
-- 1.10.x
 - 1.11.x
```
spenczar: Twirp promises to support the last two minor versions, just like Go's promise with security updates. We can't stop supporting go1.10.

cep21: > Twirp promises to support the last two minor versions

Which patch versions of the minor versions?  modules are only in the latest patch (1.9.7 not 1.9.6 for example) of older versions

cep21:
I believe you'll have to move everything to a sub directory and bump the major version number if you want to keep support for older versions of go in later versions of twirp.

marwan-at-work:
closing in favor of the discussion in #169 :)

PR #181, opened 2019-06-21
Fix use of rand.Intn

bancron:
The example uses rand.Intn(4) on a slice of len 5, meaning the last element (e.g. "blue") can never be returned. rand.Intn should be called with the len of the slice.

### DIFF
--- a/example/cmd/server/main.go
+++ b/example/cmd/server/main.go
@@ -33,7 +33,7 @@ func (h *randomHaberdasher) MakeHat(ctx context.Context, size *example.Size) (*e
 	}
 	colors := []string{"white", "black", "brown", "red", "blue"}
 	names := []string{"bowler", "baseball cap", "top hat", "derby"}
-	return &haberdasher.Hat{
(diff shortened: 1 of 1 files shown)


example/cmd/server/main.go
```diff
@@ -31,10 +31,12 @@ func (h *randomHaberdasher) MakeHat(ctx context.Context, size *example.Size) (*e
 	if size.Inches <= 0 {
 		return nil, twirp.InvalidArgumentError("Inches", "I can't make a hat that small!")
 	}
-	return &example.Hat{
+	colors := []string{"white", "black", "brown", "red", "blue"}
+	names := []string{"bowler", "baseball cap", "top hat", "derby"}
+	return &haberdasher.Hat{
```
spenczar: Please fix this to `example.Hat`, not `haberdasher.Hat`. The current code does not compile.

bancron: Sorry about that; fixed.

PR #182, opened 2019-06-21
Improve handling of malformed requests

jeff-garneau:
*Description of changes:*
- Adds (tested) specific handling for malformed JSON/protobuf requests. This can happen easily if a hand-crafted client is sending a mistyped attribute, or in general if there is bad behavior around the protobuf message definitions (i.e. changing a type on a message without moving to a new attribute number).  Not sure if the message returned here ("failed to parse request json"/"failed to parse request proto") is ideal.

Currently those requests would return 500-class errors, after this change they return 400s that help the user debug the issue.


protoc-gen-twirp/generator.go
```diff
@@ -644,6 +644,14 @@ func (t *twirp) generateUtils() {
 	t.P(`func (e *internalWithCause) WithMeta(key string, val string) `, t.pkgs["twirp"], `.Error { return e }`)
 	t.P()
 
+	t.P(`// badRequestError is used when the twirp server cannot unmarshal a request`)
+	t.P(`func badRequestError(cause error, msg string) `, t.pkgs["twirp"], `.Error {`)
+	t.P(`	err := `, t.pkgs["twirp"], `.NewError(`, t.pkgs["twirp"], `.InvalidArgument, msg)`)
+	t.P(`	err = err.WithMeta("twirp_invalid_request", cause.Error())`)
+	t.P(`	return err`)
+	t.P(`}`)
```
marioizquierdo: Maybe we should use the error code `BadRoute` instead of `InvalidArgument`.

Granted, this will result in a 404 instead of a 400 error, which may be confusing on the metrics. But it fits better with the error codes philosophy that `BadRoute` is only returned by the generated code, while `NotFound`, `InvalidArgument` and `Unimplemented` are returned by the handler code. Using generated Twirp clients should never trigger this error (that's why it used to be an internal error), but `BadRoute` is expected when doing manual API calls (e.g. curl, or custom made clients).

protoc-gen-twirp/generator.go
```diff
@@ -644,6 +644,14 @@ func (t *twirp) generateUtils() {
 	t.P(`func (e *internalWithCause) WithMeta(key string, val string) `, t.pkgs["twirp"], `.Error { return e }`)
 	t.P()
 
+	t.P(`// badRequestError is used when the twirp server cannot unmarshal a request`)
+	t.P(`func badRequestError(cause error, msg string) `, t.pkgs["twirp"], `.Error {`)
+	t.P(`	err := `, t.pkgs["twirp"], `.NewError(`, t.pkgs["twirp"], `.InvalidArgument, msg)`)
+	t.P(`	err = err.WithMeta("twirp_invalid_request", cause.Error())`)
```
marioizquierdo: I think being able to see the error is important to know what character was misspelled, so it is easy to correct. But we should double check if exposing the Go error can introduce security risks ?

protoc-gen-twirp/generator.go
spenczar: I agree that `invalid_argument` feels wrong, here. It's not that an argument was invalid for the application, it's that the request was just garbled. 

I think we would need a new error code, really. The gRPC error codes which inspired Twirp's use `INTERNAL` for this case. If that's the wrong code - which I'm not totally convinced of - then we should be descriptive. I don't think BadRoute is right, either; its name implies a routing issue, which this isn't.

spenczar:
This is a change to the spec. https://github.com/twitchtv/twirp/blob/master/docs/spec_v5.md#error-codes says "Twirp specific issues like wire and serialization problems are also reported as "internal" errors."

@jeff-garneau, can you say more about why this change is important or useful? It sounds like you've found that 5xx errors lead to service owners looking in their service instead of looking for bugs in clients, is that about right?

protoc-gen-twirp/generator.go
jeff-garneau: I agree that a new error code would be the most correct option here, though I was concerned about possible backwards-compatibility issues with old clients receiving an unrecognized error code. It seems like at least the current version of the generated go code transforms unrecognized errors into InternalErrors, so we're safe there though it would be worth checking old versions to see how this would be handled on older clients. I'm not familiar with the way the Python clients are generally used but we don't seem to provide a well-defined list of errors to Python client users so it seems up to the users to handle this case defensively.

protoc-gen-twirp/generator.go
jeff-garneau: I agree, this might be leaking more information about the underlying system (i.e. json/protobuf decoding version) than is really necessary. A generic message along the lines of `the [protobuf/json] message could not be decoded` points the user in the right direction without posing a security risk.

jeff-garneau:
In practice, this has the issue of allowing an attacker to send malformed messages resulting in status 500 responses. Automated systems may then take action based on an elevated number of 500-class responses, erroneously assuming that some system within the server is fundamentally broken. I'm primarily familiar with Elastic Beanstalk, which will mark hosts that respond with a high number of 500s as unhealthy and attempt to remove them from the loadbalancing pool, terminate and replace them. The terminations and replacements can result in denial-of-service to the system, in addition to the unnecessary alerts and actions taking place.

I'm not that familiar with other monitoring/service management systems, but in general it seems that they draw on the RFC2616 HTTP spec's distinction between 400s ("Client Error - The request contains bad syntax or cannot be fulfilled") and 500s ("Server Error - The server failed to fulfill an apparently valid request").

Basically - it's not just that the error is misleading and wastes developer time, it also has real knock-on effects that can cause real issues.

In a more general sense, I don't understand the reasoning behind that aspect of Twirp's spec, or perhaps it is being misapplied here. "Wire and serialization problems" are common to pretty much every internet service, not unique to Twirp. To me, it seems to come from the assumption that the calling client will always be the same client code that was generated alongside the server code, which is not necessarily true even if both the client and server are controlled by the same person. If that is the expectation, I feel like this should be enforced with some kind of signature that encodes the version on code generation. Instead, Twirp is deliberately opening the server to non-generated clients by including what would otherwise be a completely unnecessary JSON frontend.

marioizquierdo:
What if?

When the request is malformed, we return a twirp error `internal`, but instead of using HTTP code 500 we use code 400?

That would complicate the twirp-to-http-error-code a little bit, but it would keep the spec intact. From the point of view of bad Twirp clients, that is an internal error, from the point of view of HTTP middleware and bad hand-made requests, that is a 400 error 🤷‍♂

protoc-gen-twirp/generator.go
spenczar: I've checked a few clients. Python [should be fine](https://github.com/twitchtv/twirp/blob/337e90237d72193bf7f9fa387b5b9946436b7733/protoc-gen-twirp_python/main.go#L97).
 - [github.com/twitchtv/twirp-ruby](https://github.com/twitchtv/twirp-ruby/blob/025101ed99350d2d8995b2acf4a68c86207c3426/lib/twirp/client.rb#L67-L69) will return an Internal error
 - [github.com/thechriswalker/twirp-js](https://github.com/thechriswalker/twirp-js/blob/master/src/client.js#L67-L73) will accept anything
 - [github.com/larrymyers/protoc-gen-twirp_typescript](https://github.com/thechriswalker/twirp-js/blob/master/src/client.js#L67-L73) will accept anything
 - [github.com/twirphp/twirp](https://github.com/twirphp/twirp/blob/a76d5f30be04d4f872722cddfc4639432311996a/example/client.php#L16) will accept anything

spenczar:
@jeff-garneau thanks for the detailed justification - I'm totally convinced. Now it's really about whether we can get away with adding a new error code. Something like `bad_encoding` or `bad_serialization` or `malformed`. I lean most toward `malformed`; it's descriptive, but flexible enough in case we have similar issues in the future. It gets status code 400.

Old clients will continue to call it an Internal error, but at least they'll see it as a 4xx. More importantly (as you say) the server will not appear to be unhealthy.

>>> In a more general sense, I don't understand the reasoning behind that aspect of Twirp's spec, or perhaps it is being misapplied here. "Wire and serialization problems" are common to pretty much every internet service, not unique to Twirp. To me, it seems to come from the assumption that the calling client will always be the same client code that was generated alongside the server code, which is not necessarily true even if both the client and server are controlled by the same person. If that is the expectation, I feel like this should be enforced with some kind of signature that encodes the version on code generation. Instead, Twirp is deliberately opening the server to non-generated clients by including what would otherwise be a completely unnecessary JSON frontend.

You understand it right: it comes from an assumption that Twirp will be used through code generation, not hand-written clients. This is a slip-up that comes from Twirp's gRPC heritage. You're right that we need to support non-generated clients, really.

Signatures are an interesting idea - the Go code sends a `Twirp-Version` header today but it is never inspected for anything. I don't think I'd want hard signatures because they make it hard to update clients independently of servers.

@marioizquierdo, I think the simple mapping of errors to status codes is too valuable to give up. I'd rather add an error code.

---

@jeff-garneau what do you think of `malformed` as the new error code? Do you think it's descriptive enough?

jeff-garneau:
That makes sense to me. I'll add the `malformed` code and return that. I should have an update out soon that addresses the other concerns as well.

jeff-garneau:
Updated the docs

marioizquierdo:
It would be nice to add issues on other implementations of Twirp to adopt, or at least consider, the new error code `Malformed`. I already added one on twirp-ruby so I'll address it later: https://github.com/twitchtv/twirp-ruby/issues/39

amitmahbubani:
@spenczar JFYI. This PR introduced a breaking change. Breaks semver :(

Issue #161, opened 2019-03-17
Proposal: Expose method to write error to http.ResponseWriter


ofpiyush:
We need to check for some headers in a HTTP middleware and would love to return twirp style errors.

Right now we've copied code from generated code. That may or may not change between twirp spec versions and end up breaking our code

```go
type twerrJSON struct {
	Code string            `json:"code"`
	Msg  string            `json:"msg"`
	Meta map[string]string `json:"meta,omitempty"`
}

func Middleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		myVar := r.Header.Get("MY_KEY")
		if myVar == "" {
			err := twirp.InternalError("Could not find MY_KEY in request")
			w.Header().Set("Content-Type", "application/json")
			statusCode := twirp.ServerHTTPStatusFromErrorCode(err.Code())
			w.WriteHeader(statusCode)
			twirpErr := &twerrJSON{
				Code: string(err.Code()),
				Msg:  err.Msg(),
				Meta: err.MetaMap(),
			}
			json.NewEncoder(w).Encode(twirpErr)
			return
		}
		// More code here
	})
}
```

If the supported method exists, it can be as simple as
```go
func Middleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		myVar := r.Header.Get("MY_KEY")
		if myVar == "" {
			err := twirp.InternalError("Could not find MY_KEY in request")
			twirp.RespondWithError(w, err)
			return
		}
		// More code here
	})
}
```

spenczar:
Yeah, this has been a fairly frequent source of pain.

One reason we have avoided doing this is that we fiddled with the encoding of errors in early (pre-public) versions of Twirp. By keeping it in the generated code, it was easier to make changes that broke compatibility, since all we needed was for the generated client and server to agree on an encoding, not anyone else.

As Twirp has matured into having a real specification for error encoding, that argument is less persuasive; we can't just break error encoding. I think this is probably worth doing. It would improve a lot of code.

spenczar:
This is fixed in #192

PR #192, opened 2019-11-01
Adds WriteError func for writing Twirp errors to http.ResponseWriter

### LINKED ISSUE #186: Add WriteError to the twirp package
`WriteError` is great for standardizing twirp errors.

Proposal is to add `WriteError`-like functionality to `twitchtv/twirp` so we can get structured errors in http middlewares outside twirp services. 

> users would be able to access WriteError without importing an rpc service. Especially since twitchtv/twirp is probably already being imported to construct an error

1. add a `WriteError`-like implementation to `twitchtv/twirp` itself

2. add a method (not backwards-compatible) to `Error` interface that's "opposite" of `Msg() string`, i.e., returns a _non_ human-readable, structured messages describing the error.

Thoughts?

---

We now have multiple twirp services mounted to our router, but a single middleware stack for these specific services. So which `<service>.WriteError` would we use? It probably doesn't matter and creates an unnecessary import.

mfridman:
Fixes #186 #161 

This changeset copies the twirp service implementation of `WriteError` to the core `twitchtv/twirp` package. 

Note, all mention of "hooks" and underlying use of context has been removed. 

---

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -41,7 +41,12 @@
 //
 package twirp
 
-import "fmt"
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+)
 
 // Error represents an error in a Twirp service call.
 type Error interface {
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -346,3 +351,70 @@ func (e *wrappedErr) WithMeta(key string, val string) Error {
...
+	if writeErr != nil {
+		// We have two options here. We could log the error, or just silently
+		// ignore the error.
+		//
+		// Logging is unacceptable because we don't have a user-controlled
+		// logger; writing out to stderr without permission is too rude.
+		//
+		// Silently ignoring the error is our least-bad option. It's highly
+		// likely that the connection is broken and the original 'err' says
+		// so anyway.
```
ofpiyush: or return the error to the caller and let them log/ignore?

spenczar: Ah! You're completely right that this is a very different situation from the generated code. This should most certainly be able to return an error.

It's unfortunate that this would be a breaking change and require a major version bump. This is why public API is hard. Ouch.

mfridman: Indeed this could return an error, and probably should within this pkg.

But on the flip-side, it would no longer be a drop-in replacement to the generated version of `WriteError`.

In our case we were importing stub packages in various places outside the twirp stack, so it made sense to have it this way.

But it wouldn't be too hard to refactor with the returned error (we'd probably drop it anyways).

Maybe in a v6 we could add a returned error, or in v5 add a new function that returns the underlying error.

ofpiyush: It would technically break the public API, but would not do so in a code-breaking way.

Since this function doesn't return anything, all existing code would look like

```
twirp.WriteError(resp, err)
```

This would not break if you started returning errors. i.e. you can get away with adding it in a minor version.

If that feels too close to the edge of the dark side.

> v5 add a new function that returns the underlying error.

💯 
I do not claim to know how to cleanly fix a public API. Probably something like this work?

- Add another method to return error
- Mark the current one deprecated
- In V6, drop the current method

spenczar: I like this discussion - let's move it to an issue though, since I don't think anybody else is seeing this :) @ofpiyush, can you make an issue?

PR #194, opened 2019-11-13
[DO NOT MERGE] RFC: Client hooks redux

ccmtaylor:
*Issue #, if available:* #145 

*Description of changes:*

this modifies #164 slightly to move the HTTPClient type back into the generated code.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #198, opened 2019-11-27
Add support for Client Hooks

### LINKED ISSUE #145: Proposal: Client-Side Hooks in Go

iheanyi:
*Closes #145*

This pull request adds more fleshed out support for Client Hooks (from #164 and #194) as discussed in #145. It adds support for a few basic hooks:

* `RequestPrepared` - Called after a request object is instantiated but before it is sent to a server.
* `ResponseReceived` - Called after a request is successfully sent.
* `Error` - Called when there is an error sending a request.


client.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018 Twitch Interactive, Inc.  All Rights Reserved.
```
ccmtaylor: maybe rename to `client_hooks.go`/`client_hooks_test.go` to be consistent with the server-side filenames?

client.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type ClientHooks struct {
+	// RequestPrepared is called as soon as a request has been created and before it has been sent
+	// to the Twirp server.
+	RequestPrepared func(context.Context, *http.Request) (context.Context, error)
+
+	// RequestFinished is called after a request has finished sending. Since this is terminal, the context is
+	// not returned.
+	RequestFinished func(context.Context)
```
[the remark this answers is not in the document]
iheanyi: Yeah, I agree, let's rename it. Second question: do we want to pass `res *http.Response` into this function? 🤔Maybe overkill and could be a footgun for people reading the body and stuff in hooks.

client.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Error hook is called whenever an error occurs during the sending of a request. The Error is passed
+	// as an argument to the hook.
+	Error func(context.Context, Error) context.Context
```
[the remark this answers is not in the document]
iheanyi: Yeah, that's what I was thinking as well. If we tweak it further to use that returned value we can always back in.

client.go
iheanyi: Ah, touché.

iheanyi: I only ended up renaming `client.go` since `client_test.go` is in the `twirptest` package and it's a holistic test around the client, which hooks are a part of. Fixed in https://github.com/twitchtv/twirp/pull/198/commits/ee862b21bd2a8b1cc8eae533b8c1ba7fe74f19c4.

client.go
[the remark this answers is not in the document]
iheanyi: Fixed in https://github.com/twitchtv/twirp/pull/198/commits/89b596bf0d92e9d22699c20da6fa255dffe219e5.

client.go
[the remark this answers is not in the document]
iheanyi: Fixed in https://github.com/twitchtv/twirp/pull/198/commits/33e3c737ea4b131839e5753a306f73babc28733e.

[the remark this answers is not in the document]
ccmtaylor: good question, I'm not sure. I could imagine it might be convenient to poke at response headers or status codes, but otoh HTTP-level instrumentation should probably go in a RoundTripper on the HTTPClient.

internal/twirptest/client_test.go
```diff
@@ -257,8 +257,94 @@ func TestClientRedirectError(t *testing.T) {
 	})
 }
 
+func TestClientWithHooks(t *testing.T) {
+	requestPreparedCalled := false
+	requestFinishedCalled := false
```
ccmtaylor: `%s/equestFinished/esponseReceived/g` in this file :)

iheanyi: Ah, good catch. 🤦‍♂

iheanyi:
All right @ccmtaylor, addressed your feedback and a thought came up while doing so. We may want to return the `ctx` for `Error` after we implement `ChainHooks` so we can pass the updated `ctx` to the `Error` hook. But then again, it may be okay not to if `Error` hook is one of two terminal  hooks. 🤔

ccmtaylor:
I implemented ClientChainHooks in https://github.com/iheanyi/twirp/pull/1, you can pull it into this PR by merging that one.

re returning the context: I think it makes sense for hooks on the _outgoing_ path, e.g. to allow setting deadlines/timeouts. For hooks handling incoming responses, I can't think of a good use-case, since the modified context gets discarded (from the twirp-end-user perspective) when the request is done. It'd feel a bit icky if one hook depends on changes that an earlier hook makes to the context. OTOH, I don't see a strong reason not to make it possible, but I'm leaning against it slightly.

iheanyi:
LGTM @ccmtaylor, thanks for landing `ChainClientHooks`. Anything else in the PR that may have caught your eye? If not, would love a review from @spenczar so we can iterate on it if necessary.

ccmtaylor:
nope, LGTM, too :+1:

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ClientOptions encapsulate the configurable parameters on a Twirp client.
+type ClientOptions struct {
+	Hooks *ClientHooks
+}
+
+// ClientHooks is a container for callbacks that can instrument a
+// Twirp-generated client. These callbacks all accept a context and some return
+// a context. They can use this to add to the contxt, appending values or
+// deadlilns to it.
```
spenczar:
```suggestion
// deadlines to it.
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+type ClientOption func(*ClientOptions)
+
+// ClientOptions encapsulate the configurable parameters on a Twirp client.
+type ClientOptions struct {
+	Hooks *ClientHooks
+}
+
+// ClientHooks is a container for callbacks that can instrument a
+// Twirp-generated client. These callbacks all accept a context and some return
+// a context. They can use this to add to the contxt, appending values or
```
spenczar:
```suggestion
// a context. They can use this to add to the context, appending values or
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+// deadlilns to it.
+//
+// The RequestPrepared hook is special because it can return errors. If it
+// returns non-nil error, handling for that request will be stopped at that
+// point. The Error hook will then be triggered.
+//
+// The RequestPrepaared hook will always be called first and wlil be called for
```
spenczar:
```suggestion
// The RequestPrepaared hook will always be called first and will be called for
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+// every outgoinig request from the Twirp client. The last hook to be called
```
spenczar:
```suggestion
// each outgoing request from the Twirp client. The last hook to be called
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+// will either be Error or ResponseReceived, so be sure to handle both case in
```
spenczar:
```suggestion
// will either be Error or ResponseReceived, so be sure to handle both cases in
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+	// being returned from the request.
+	ResponseReceived func(context.Context)
+
+	// Error hook is called whenever an error occurs during the sending of a request. The Error is passed
+	// as an argument to the hook.
+	Error func(context.Context, Error)
+}
+
+// DefaultClientOptions instantiates ClientOptions with their default options.
+func DefaultClientOptions() ClientOptions {
```
spenczar: Please remove this function. We can document that a zero-value ClientOptions provides default behavior, at least for now.

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+// will either be Error or ResponseReceived, so be sure to handle both case in
+// your hooks.
+type ClientHooks struct {
+	// RequestPrepared is called as soon as a request has been created and before it has been sent
+	// to the Twirp server.
+	RequestPrepared func(context.Context, *http.Request) (context.Context, error)
+
+	// ResponseReceived is called after a request has finished sending. Since this is terminal, the context is
+	// not returned. ResponseReceived will not be called in the case of an error
+	// being returned from the request.
```
spenczar: Please adjust this to fit line lengths, to this:
```
	// ResponseReceived is called after a request has finished sending. Since this
	// is terminal, the context is not returned. ResponseReceived will not be
	// called in the case of an error being returned from the request.
```

client_hooks.go
```diff
@@ -0,0 +1,111 @@
...
+	ResponseReceived func(context.Context)
+
+	// Error hook is called whenever an error occurs during the sending of a request. The Error is passed
+	// as an argument to the hook.
```
spenczar: Please adjust this to fit line lengths too:
```
		// Error hook is called whenever an error occurs during the sending of a
		// request. The Error is passed as an argument to the hook.
```

example/service.twirp.go
```diff
@@ -45,24 +45,31 @@ type Haberdasher interface {
...
 }
 
 // NewHaberdasherProtobufClient creates a Protobuf client that implements the Haberdasher interface.
 // It communicates using Protobuf and can be configured with a custom HTTPClient.
-func NewHaberdasherProtobufClient(addr string, client HTTPClient) Haberdasher {
+func NewHaberdasherProtobufClient(addr string, client HTTPClient, opt ...twirp.ClientOption) Haberdasher {
+	var httpClient HTTPClient = client
+	if c, ok := client.(*http.Client); ok {
+		httpClient = withoutRedirects(c)
+	}
```
spenczar: The `httpClient` variable is unnecessary. You can do `client = withoutRedirects(c)` inside the braces.

example/service.twirp.go
```diff
@@ -45,24 +45,31 @@ type Haberdasher interface {
 type haberdasherProtobufClient struct {
 	client HTTPClient
 	urls   [1]string
+	opts   twirp.ClientOptions
 }
 
 // NewHaberdasherProtobufClient creates a Protobuf client that implements the Haberdasher interface.
 // It communicates using Protobuf and can be configured with a custom HTTPClient.
-func NewHaberdasherProtobufClient(addr string, client HTTPClient) Haberdasher {
+func NewHaberdasherProtobufClient(addr string, client HTTPClient, opt ...twirp.ClientOption) Haberdasher {
```
spenczar: Please name the argument `opts` instead of `opt`, to match its name within the struct (and the fact that it is variadic).

example/service.twirp.go
```diff
@@ -85,24 +100,31 @@ func (c *haberdasherProtobufClient) MakeHat(ctx context.Context, in *Size) (*Hat
...
+	var httpClient HTTPClient = client
+	if c, ok := client.(*http.Client); ok {
+		httpClient = withoutRedirects(c)
+	}
+
+	opts := twirp.DefaultClientOptions()
+	for _, o := range opt {
+		o(&opts)
+	}
+
```
spenczar: All the comments on the ProtobufClient constructor apply here, too.

internal/twirptest/client_test.go
```diff
@@ -257,8 +257,94 @@ func TestClientRedirectError(t *testing.T) {
...
+	var resp *Hat
+	var err error
+
+	// Test proto success
+	resp, err = protoCli.MakeHat(ctx, &Size{Inches: 1})
+	if err != nil {
+		t.Fatalf("Proto client method returned unexpected error: %s", err)
+	}
+	if resp == nil {
+		t.Fatalf("Proto client method expected to return non-nil response, but it is nil")
```
spenczar: Use `t.Fatal`, since there is no formatted argument. The same applies to a bunch of the calls below.

example/service.twirp.go
```diff
@@ -678,6 +708,12 @@ func doProtobufRequest(ctx context.Context, client HTTPClient, url string, in, o
 	if err != nil {
 		return wrapInternal(err, "could not build request")
 	}
+	ctx, err = callClientRequestPrepared(ctx, hooks, req)
+	if err != nil {
+		return wrapInternal(err, "failed to call RequestPrepared hook")
```
spenczar: This error text is a little misleading. It isn't necessarily that we failed to call `RequestPrepared` - it's that it returned an error. That's probably because the hook errored - the programmer _wanted_ the error.

Wrapping the error seems a little odd here, too. I'd expect that the error value returned by RequestPrepared is exactly what I'd received in an Error hook, not a wrapped version of it. I'm interested in alternatives to what you have here (or thoughts on whether the API seems correct as-is).

internal/twirptest/client_test.go
```diff
@@ -257,8 +257,94 @@ func TestClientRedirectError(t *testing.T) {
...
+	}
+
+	if requestPreparedCalled == false || responseReceivedCalled == false {
```
spenczar: Please evaluate these cases individually. Use `t.Error` instead of `t.Fatal`, so that each case gets tested rather than exiting early, and also test that `errorCalled == false`.

internal/twirptest/client_test.go
```diff
@@ -257,8 +257,94 @@ func TestClientRedirectError(t *testing.T) {
...
+	// Test proto failure
+	resp, err = protoCli.MakeHat(ctx, &Size{Inches: 666})
+	if err == nil {
+		t.Fatalf("Proto client method expected to fail, but error is nil")
+	}
+	if resp != nil {
+		t.Fatalf("Proto client method expected to return nil response on error, but returned non-nil")
+	}
+
+	if errorCalled == false {
```
spenczar: I think you need to reset the test state before this check. That might mean a refactor of this test in general.

example/service.twirp.go
iheanyi: @spenczar That's a solid point, I forgot that we're going to make the error into an `InternalError` from the caller anyways, so this is unnecessary. I'll remove it and just return the `err` itself.

iheanyi:
@spenczar Thanks for the first round of review! I've addressed your feedback and would love another lookover when you get time. Some particularly big changes to note:

* Not wrapping errors returned from calling the request prepared hook.
* Converting `TestClientWithHooks` function to use test tables and explicitly test the state of both generated clients, also added in a test for covering the returned error from the `RequestPrepared` hook.

Let me know if there's anything else you'd like me to address or if I missed anything. Thanks in advance!

client_hooks.go
```diff
@@ -0,0 +1,104 @@
...
+func ChainClientHooks(hooks ...*ClientHooks) *ClientHooks {
+	if len(hooks) == 0 {
+		return nil
+	}
+	if len(hooks) == 1 {
+		return hooks[0]
+	}
+	return &ClientHooks{
+		RequestPrepared: func(ctx context.Context, req *http.Request) (context.Context, error) {
+			var err error
```
bufdev: nit: no need to declare this in top-level function scope, you can instead do:

```go
			for _, h := range hooks {
				if h != nil && h.RequestPrepared != nil {\
                                        var err error
					ctx, err = h.RequestPrepared(ctx, req)
					if err != nil {
						return ctx, err
					}
				}
			}
			return ctx, nil
```

client_hooks.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"net/http"
+)
+
+// ClientOption is a functional option for extending a Twirp client.
+type ClientOption func(*ClientOptions)
+
+// ClientOptions encapsulate the configurable parameters on a Twirp client.
+type ClientOptions struct {
+	Hooks *ClientHooks
```
bufdev: I'm assuming the ServerHooks are implemented in the same way, but you could instead:

- Make this a private `hooks *ClientHooks`
- Make `WithClientHooks` implement the logic in `ChainClientHooks` by automatically chaining if `hooks != nil`
- Have `func (h *ClientHooks) Hooks() *ClientHooks` that the generated code calls

This would simplify the API a bit

client_hooks.go
iheanyi: Good call, will change!

client_hooks.go
iheanyi: Ahhhhh, this is a good idea. I'll give this a shot. For the third bullet point, did you mean: `func (o *ClientOptions) Hooks() *ClientHooks`?

iheanyi:
@spenczar Pushed up some suggestions that simplified the API a little bit, happy to revert if you feel like they are too different from the server implementation or anything else.

client_hooks.go
iheanyi: Addressed.

client_hooks.go
```diff
@@ -0,0 +1,107 @@
...
+
+// WithClientHooks sets the *ClientHooks for a Twirp client which chains the
+// callbacks in each of the constituent hooks passed in. Each hook function will
+// be called in the order of the ClientHooks values passed in.
+
+// For the erroring hook, RequestPrepared, any returned errors prevent
+// processing by later hooks.
+func WithClientHooks(hooks ...*ClientHooks) ClientOption {
+	return func(o *ClientOptions) {
+		o.hooks = chainClientHooks(hooks...)
```
bufdev: Two issues:

1. This should be `o.hooks = chainClientHooks(append(o.hooks, hooks...)...)`
2. I would chain when calling `Hooks()`, once (i.e. include a `sync.Once` as a private field of your `ClientOptions`. If you stick with the current implementation and update per (1), this means that every new call to `WithClientHooks` results in a recursive method call courtesy of `chainClientHooks).

```go
type ClientOptions struct {
  allHooks []*ClientHooks
  hooks *ClientHooks
  once sync.Once
}

func WithClientHooks(hooks ...*ClientHooks) ClientOption {
  return func(o *ClientOptions) {
    o.allHooks = append(o.allHooks, hooks...)
  }
}

func (o *ClientOptions) Hooks() *ClientHooks {
  o.once.Do(o.initHooks)
  return o.hooks
}

func (o *ClientHooks) initHooks() {
  o.hooks = chainClientHooks(o.allHooks...)
}
```

iheanyi: Hmmmmm, I think with these changes, this gets a little bit more complicated than what we had before, don't you think?

iheanyi: I'm also a little bit confused, `WithClientHooks` will be called at most once, right? For example, `NewProtobufClient(addr, client, twirp.WithClientHooks(hook1, hook2)`. On the other hand, are you saying to make it just in case they do `twirp.WithClientHooks(hook1), twirp.WithClientHooks(hook2)`?

bufdev: It allows both use cases:

- `twirp.WithClientHooks(hook1, hook2, hook3)`
- `twirp.WithClientHooks(hook1), twirp.WithClientHooks(hook2, hook3)`

iheanyi:
Reverted the variadic changes until there's more feedback, just because I felt like internally it's getting a bit more complicated, happy to change back but will wait until Spencer chimes in.

spenczar:
Yeah, I prefer the `ChainClientHooks` function over needing to fiddle with stuff like `sync.Once`. the Chain function is clear and simple, and matches the ServerHooks API, which we can't change without breaking compatibility harshly.

I'll take a look at the rest of the code soon.

iheanyi:
Agreed @spenczar, looking forward to landing this!

client_hooks.go
```diff
@@ -0,0 +1,104 @@
...
+// ClientHooks is a container for callbacks that can instrument a
+// Twirp-generated client. These callbacks all accept a context and some return
+// a context. They can use this to add to the context, appending values or
+// deadlines to it.
+//
+// The RequestPrepared hook is special because it can return errors. If it
+// returns non-nil error, handling for that request will be stopped at that
+// point. The Error hook will then be triggered.
+//
+// The RequestPrepaared hook will always be called first and will be called for
```
spenczar: Typo: `RequestPrepaared`

PR #199, opened 2019-11-27
Upgrade protobuf version

### LINKED ISSUE #197: Upgrade Protobuf version
I've noticed that the Protobuf version has gotten pretty out of date. Protobuf 3.5.1 was released in December of 2017, whereas you have later versions that support more things such as support for newer version of languages such as Python. I think it would be helpful l to update the version of the Protobuf dependency used by Twirp.

iheanyi:
*Closes #197*

This pull request upgrades the protobuf version to be 3.10.1, which is the latest version in Homebrew as of 11/27/2019 rather than using 3.7.1, which is 2 years old. Additionally, it recompiles the Protobuf files for Python to be updated for this new version.


gorzell:
It looks like 3.11.0 was released fairly recently: https://github.com/protocolbuffers/protobuf/releases/latest. Any reason not to just go all the way up to that?

iheanyi:
@gorzell @spenczar Good call, I've updated the protobuf version to 3.11.0 and also regenerated the various clients as well. Let me know what you think once you take a lookover.

spenczar:
Thanks!

Issue #196, opened 2019-11-26
option paths=source_relative:. causes import cycle loop


strobil:
Command used for generating *twirp.go

> protoc -I /usr/local/include/ \
>   -I $GOPATH/src \
>   -I $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
>   -I . \
>   -I tv_service \
>   --go_out=plugins=grpc,paths=source_relative:. \
>   --grpc-gateway_out=logtostderr=true,grpc_api_configuration=tv_service/tv_service.yaml,paths=source_relative:. \
>   --twirp_out=paths=source_relative:. \
>   tv_service/tv_service.proto

When i am trying to build project this error occurs:

> import cycle not allowed
> package gitlab.sweet.tv/go/tv_service
> 	imports gitlab.sweet.tv/proto/tv_service
> 	imports gitlab.sweet.tv/proto/tv_service

The reason is in generated file which contains this line, but it should not, because the package name is identical in both cases (tv_service)

`import tv_service8 "gitlab.sweet.tv/proto/tv_service"`

If i remove paths=source_relative:. the generated file does not contain this import line.

spenczar:
This is hard to debug without more detail. Can you share the `.proto` file you are using, or a stripped-down version of it that minimally reproduces your error?

strobil:
@spenczar Yes, sure.

Untar these file and run go_generate.sh

[proto2.tar.gz](https://github.com/twitchtv/twirp/files/3917489/proto2.tar.gz)

spenczar:
Thanks, I'm able to reproduce your error. I'll isolate it into a test case and work on a fix.

spenczar:
@strobil can you test the branch 'fix_local_proto_imports' (https://github.com/twitchtv/twirp/pull/202) to confirm that it fixes your full build? It works for me locally.

strobil:
@spenczar, how i can install a specific version of protoc-gen-twirp?

spenczar:
@strobil Something like this:
```
export GOPATH=$(mktemp -d)
mkdir -p $GOPATH/src/github.com/twitchtv/twirp
git clone https://github.com/twitchtv/twirp.git $GOPATH/src/github.com/twitchtv/twirp
cd $GOPATH/src/github.com/twitchtv/twirp
git checkout -b fix_local_proto_imports origin/fix_local_proto_imports
[...]

PR #202, opened 2019-12-03
Fix local proto imports

spenczar:
*Issue #, if available:*
#196 

*Description of changes:*
These are some subtle changes to how we map `.proto` files to package names. It should resolve the bug in #196 by correctly identifying when an imported `.proto` file has the same Go import path as the file we're generating.


spenczar:
This is only a partial fix for #196 at present; there are multiple issues there.

spenczar:
This should now be a complete fix.

internal/twirptest/importer_local/importer_local.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package twirp.internal.twirptest.importer_local;
+option go_package = "github.com/twitchtv/twirp/internal/twirptest/importer_local";
+
+import "importer_local_msgdef.proto";
+
+service Svc {
```
dpolansky: Could you add a comment (wherever you deem appropriate) on what this package is testing for future reference?

protoc-gen-twirp/generator.go
```diff
@@ -114,22 +114,42 @@ func (t *twirp) Generate(in *plugin.CodeGeneratorRequest) *plugin.CodeGeneratorR
 	}
 	t.genPkgName = genPkgName
 
+	// We also need to figure out the fully import path of the package we're
```
dpolansky: The comments help tremendously with understanding the nuance of how package names are assigned -- thanks!

PR #205, opened 2019-12-06
Don't treat NoError as valid

aconverse:
*Description of changes:*

This indirectly fixes JSON error bodies from middle entities like API Gateway being properly detected as non-twirp errors. 

Only treat codes that convert to 400 (the lowest twirp error code) or higher as valid error codes.

See https://github.com/twitchtv/twirp/blob/v5.9.0/protoc-gen-twirp/generator.go#L538 for more context.


aconverse:
Already solved by https://github.com/twitchtv/twirp/pull/193 , my generated clients were old.

Issue #201, opened 2019-12-03
Return error from twirp.WriteError


ofpiyush:
Moving discussion from the [PR](https://github.com/twitchtv/twirp/pull/192#discussion_r344446180) as @spenczar requested.

**Recap:**

Current implementation of `twirp.WriteError` does not return any error occurred while writing the TwirpError.

<details><summary>Humor</summary><p><em>WriteError does not return the error if an error occurs while writing the error.</em></p></details>

Since the public API currently returns nothing, all existing code looks like

```go
twirp.WriteError(resp, err)
```

As no return is being processed, adding a return to this method will not break existing code.

spenczar:
You're right that it won't break most code. The only exception would be that the function's type changes, so if someone has a `var w func(*http.ResponseWriter, error) = twirp.WriteError`, that code will break. That seems pretty unusual, and the function is new enough that it might be safe to modify without any significant changes.

It's a bit of a violation of semver to make this part of a minor release, but I think it might be a forgivable one.

spenczar:
Oh, and in case it isn't obvious - yes, it should return an error if it fails to write.

PR #208, opened 2019-12-12
Return error from WriteError

ofpiyush:
*Issue #, if available:*
https://github.com/twitchtv/twirp/issues/201

*Description of changes:*

`WriteError` now returns the error it might encounter while writing.

Added tests to cover `WriteError`'s happy path and error case.

<hr>

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -14,7 +14,9 @@
 package twirp
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/http/httptest"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


ofpiyush:
PS: I ran `go fmt ./...` out of habit and it formatted `internal/gen/wrappers.go`.

Left the change in as a separate PR for just formatting felt like an overkill.

errors_test.go
```diff
@@ -49,3 +51,52 @@ func TestErrorCause(t *testing.T) {
 		t.Errorf("got wrong cause for err. have=%q, want=%q", cause, rootCause)
 	}
 }
+
+type errorResponeWriter struct {
+	*httptest.ResponseRecorder
+}
+
+func (errorResponeWriter) Write([]byte) (int, error) {
+	return 0, fmt.Errorf("this is only a test")
```
spenczar: nit: this could be `errors.New`

ofpiyush: I thought of that and then figured the file has fmr.Errorf above and went with it instead.


Sorry didn't even see this till today!


The next time I send a PR, I'll club with that 😊

PR #225, opened 2020-04-14
 feat(protoc-gen-twirp): support gzipped response with protobuf

knqyf263:
*Description of changes:*
I'm facing a problem with a large response. It is slow and also sometimes hits a limitation of response size in a cloud service. I'd like to support gzip compression. If it is acceptable, I can send another PR for JSON as well.

### DIFF
--- a/protoc-gen-twirp/generator.go
+++ b/protoc-gen-twirp/generator.go
@@ -105,6 +105,7 @@ func (t *twirp) Generate(in *plugin.CodeGeneratorRequest) *plugin.CodeGeneratorR
 	t.registerPackageName("twirp")
 	t.registerPackageName("url")
 	t.registerPackageName("fmt")
+	t.registerPackageName("gzip")
 
 	// Time to figure out package names of objects defined in protobuf. First,
 	// we'll figure out the name for the package we're generating.
@@ -203,6 +204,8 @@ func (t *twirp) generate(file *descriptor.FileDescriptorProto) *plugin.CodeGener
 		t.generateUtilImports()
 	}
 
+	t.generateConst()
+
 	// For each service, generate client stubs and server
 	for i, service := range file.Service {
 		t.generateService(file, service, i)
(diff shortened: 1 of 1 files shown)


dpolansky:
Although I agree that compression is worth exploring, this implementation cannot be merged as is. 

My primary concern here is that clients and servers will experience a breaking change in the Twirp's expected behavior if both the client and server were upgraded to include this implementation. In its current form, it seems like the client automatically opts-in to the compression behavior by implicitly specifying the request header. It's also possible that some clients are including this request header already unintentionally and may be surprised when the server's behavior changes. In my opinion, a feature like this can only be considered if it is opt-in only.

If you would like to explore this use case further, please submit a proposal as a Github issue so that the community can discuss it.

knqyf263:
Thank you for the reply, but I don't get your point. If a client doesn't include the header, a server will not compress. If the client adds the header, it means the client can decompress the response. Why is the client going to be surprised even though the client specifies the header?

protoc-gen-twirp/generator.go
```diff
@@ -468,6 +479,7 @@ func (t *twirp) generateUtils() {
 	t.P(`  }`)
 	t.P(`  req.Header.Set("Accept", contentType)`)
 	t.P(`  req.Header.Set("Content-Type", contentType)`)
+	t.P(`  req.Header.Set("Accept-Encoding", "gzip")`)
```
dpolansky, line 482: This change makes it such that all clients using this code include the `Accept-Encoding` header.

knqyf263: If a server doesn't support gzip, this header will be just ignored.

dpolansky: Right, I understand that. However, a feature like this will only be considered if it is _opt-in_, meaning that clients and servers can safely upgrade their version of Twirp without experiencing any change in how Twirp servers encode responses. 

The current implementation is not opt-in because a user may upgrade their Twirp client and experience a change in behavior. This kind of change would require a release of a new major version.

knqyf263: How about addind an option to enable compression which is disabled by default?

dpolansky: Please submit a proposal for how this feature should be implemented in Twirp as a Github issue

knqyf263: Get it

Issue #228, opened 2020-04-24
Extend error handling capabilities


flimzy:
The package's error handling could be made more flexible.  The particular problem we're facing is that error context is stripped by this package, when it converts errors to `twirp.Error`.

I think the simplest change, which should be 100% backward compatible, would be to extend the `twirp.Error` interface (and its implementations) could be expanded to support the new `errors.Unwrap` interface.  Then, of course, make it possible to wrap existing errors, as well. `InteralWithError` already has the right function signature. We'd need the same for other error types, too, so I'd suggest adding a general-purpose function, perhaps simply `WrapError(ErrorCode, error) Error`.

flimzy:
Looking more closely, I see this is closer to reality than I previously thought, as the `wrappedErr` type already exists.  I think that type could be easily extended to support the `Unwrap` method, and it would still be useful to allow wrapping errors of other types, as well.

PR #229, opened 2020-04-24
Allow wrapping any error code, not just Internal errors

flimzy:
And support the Unwrap() method in addition to Cause()

Issue #228 

*Description of changes:*

Add a function `WrapError(ErrorCode, error) Error`, which allows wrapping an error with any error code, not just Internal errors.

And add the `Unwrap()` method to the `wrappedErr` type, to be compatible with Go 1.13's new error unwrapping.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -72,19 +72,10 @@ type Error interface {
 	Error() string
 }
 
-// WrapError allows wrapping another error. It adds the underlying error's type
-// as metadata with a key of "cause", which can be useful for debugging.
-//
-// The returned error also has both Cause() and Unwrap() methods which will
-// return the original error. This can be used with the github.com/pkg/errors
-// package, or the standard `errors` package from Go 1.13 onward, to extract
-// the root cause of an error. Information about the root cause of an error is
-// lost when it is serialized, so this doesn't let a client know the exact root
-// cause of a server's error.
-func WrapError(code ErrorCode, err error) Error {
-	msg := err.Error()
-	twerr := NewError(code, msg)
-	twerr = twerr.WithMeta("cause", fmt.Sprintf("%T", err)) // to easily tell apart wrapped internal errors from explicit ones
+// WrapError allows Twirp errors to wrap other errors.
+// The wrapped error can be extracted later with (github.com/pkg/errors).Unwrap
+// or errors.Is from the standard errors package on Go 1.13+.
+func WrapError(twerr Error, err error) Error {
 	return &wrappedErr{
 		wrapper: twerr,
 		cause:   err,
@@ -135,14 +126,15 @@ func InternalError(msg string) Error {
 	return NewError(Internal, msg)
 }
 
-// InternalErrorWith is an easy way to wrap another error. It should be used in
-// the common case of an unexpected error returned from another API, but
-// sometimes it is better to build a more specific error (like with
-// NewError(Unknown, err.Error() or WrapError(Unknown err)), for example).
-//
-// See WrapError for additional details on the returned type.
+// InternalErrorWith makes an internal error, wrapping the original error and using it
+// for the error message, and with metadata "cause" with the original error type.
+// This function is used by Twirp services to wrap non-Twirp errors as internal errors.
+// The wrapped error can be extracted later with (github.com/pkg/errors).Unwrap
+// or errors.Is from the standard errors package on Go 1.13+.
 func InternalErrorWith(err error) Error {
-	return WrapError(Internal, err)
+	twerr := NewError(Internal, err.Error())
+	twerr = twerr.WithMeta("cause", fmt.Sprintf("%T", err)) // to easily tell apart wrapped internal errors from explicit ones
+	return WrapError(twerr, err)
 }
 
 // ErrorCode represents a Twirp error type.
--- a/errors.go
+++ b/errors.go
@@ -72,6 +72,25 @@ type Error interface {
 	Error() string
 }
 
+// WrapError allows wrapping another error. It adds the underlying error's type
+// as metadata with a key of "cause", which can be useful for debugging.
+//
+// The returned error also has both Cause() and Unwrap() methods which will
+// return the original error. This can be used with the github.com/pkg/errors
+// package, or the standard `errors` package from Go 1.13 onward, to extract
+// the root cause of an error. Information about the root cause of an error is
+// lost when it is serialized, so this doesn't let a client know the exact root
+// cause of a server's error.
+func WrapError(code ErrorCode, err error) Error {
+	msg := err.Error()
+	twerr := NewError(code, msg)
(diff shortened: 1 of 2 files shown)


flimzy:
It would also be nice to unwrap errors returned by the Twirp server (and possibly client, though this doesn't affect me personally).

In other words, it would be nice for a function to, do something like:

```golang
    err := twirp.NewError(twirp.NotFound, "it's missing")
   /* some time later */
   return errors.WithStack(err)
```
I think this change would belong here, in `errors.go`:

```golang
// writeError writes Twirp errors in the response.
func writeError(resp http.ResponseWriter, err error) error {
	// Non-twirp errors are wrapped as Internal (default)
	twerr, ok := err.(Error)
	if !ok {
		twerr = InternalErrorWith(err)
	}
```

But there are many subtleties to consider, so I'd rather get some feedback that this is desired before I work on a PR.

mfridman:
Our use case for `WriteError`, which calls `writeError` under the hood, is exclusively in the middleware portion, where we return Twirp-structured errors to frontend clients.

And if inside Twirp handlers one would call some variant that returns a twirp Error.

Maybe you can describe what changes you'd make to `writeError`? Is the proposal to call `pkg/errors.WithStack` by default to include a stack trace for every request?

flimzy:
Our ultimate goal is to log errors in a meaningful way, including additional context, such as (but not limited to) stack traces.

This could be done with the server `Error` hook, except that there is no way to pass error context through the twirp layer,  except with `InternalErrorWith`, which of course only works for internal errors.

So my suggestion with respect to `writeError` would be something like:

```golang
func writeError(resp http.ResponseWriter, err error) error {
	twerr = new(Error)
	// Non-twirp errors are wrapped as Internal (default)
	if !errors.As(err, twerr) {
		twerr = InternalErrorWith(err)
	}
```

As I mentioned above, though, there are many subtleties to consider here:

1. `errors.As` only works with `Unwrap()`, not `Cause()`. It's a trivial matter to write a version that handles both, though.
2. More important, simply using `errors.As` (or even `github.com/pkg/errors#Cause()`) will alter the result of `err.Error()`, by stripping some of its context.  What may be desired is something more equivalent to:

```golang
func writeError(resp http.ResponseWriter, err error) error {
	twerr = new(Error)
	// Non-twirp errors are wrapped as Internal (default)
	if !errors.As(err, twerr) {
		twerr = &wrappedErr{
			wrapper: NewError(twerr.Code(), err.Error()), // Code from the unwrapped, error, but err.Error() from the original
			cause: err,
		}
	}
```

I think there's a lot of subtlety here, so I hope my explanation of both the problem, and proposal, make sense.

marioizquierdo:
`twirp.Error` is meant to be a value sent through the network, not an actual error to be passed around in the service. The case for internal errors is different because that is the "default" wrapper when any other error is returned, in the service, or when a network issue happens on the client, in which case it is important to unwrap to inspect the reason for that error.

I'm curious about what other use cases you think may be solved by allowing to wrap errors with codes other than internal. Or maybe there's anything that could be explained better on the docs?

flimzy:
The main case we care about is logging error details. This is why we want the complete error to be accessible in the `Error` hook. This can be useful for virtually any error, not just "internal" errors.

The alternative is to ensure that we log all details before converting to a `twirp.Error`, which is cumbersome, and error-prone.

Example:

```golang
func SomeHandler(...) (..., error) {
    data, err := ProcessData(/* ... */)
    if err != nil {
        return twirp.NewError(twirp.Malformed, err.Error())
    }
}
```

If `ProcessData` adds additional context to the error, such as a [stack trace](https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper),  or any other arbitrary data about the context of the error, the `twirp.NewError` method discards it.  By wrapping, it would be possible to extract that info within the `Error` hook for logging purposes.

The only other options are:

1. To handle all logging before returning a twirp error. As alluded to above, this is cumbersome and error prone, because it's so easy to forget to do. It also requires code that appears to violate the [handle errors once](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) principle, by both logging _and_ returning the same error.

```golang
func SomeHandler(...) (..., error) {
    data, err := ProcessData(/* ... */)
    if err != nil {
        logErrorWithContext(err)
        return twirp.NewError(twirp.Malformed, err.Error())
    }
}
```
2. Stop using `twirp.NewError` entirely, and use a custom error type that handles the error wrapping for us.  This solves the handle-errors-once problem, but is still error prone, because it's so easy to use `twirp.NewError` instead.

marioizquierdo:
Gotcha. I see how this can be useful.

I would suggest using a slightly different implementation if that makes sense. Instead of adding the meta "cause" to all wrapped errors (which could be surprising behavior in the client), we should just allow to wrap any twirp error with an original error:

```go
// WrapError allows Twirp errors to wrap other errors.
// The wrapped error can be extracted later with (github.com/pkg/errors).Unwrap 
// or errors.Is from the standard errors package on Go 1.13+.
func WrapError(twerr Error, err error) Error {
	return &wrappedErr{
		wrapper: twerr,
		cause:   err,
	}
}

// InternalErrorWith makes an internal error, wrapping the original error and using it for the
// error message. It adds the wrapped error's type as metadata with the key: "cause".
// This function is used by Twirp services to wrap non-Twirp errors as internal errors.
// The wrapped error can be extracted later, usually on middleware, with
// (github.com/pkg/errors).Unwrap or errors.Is from the standard errors package on Go 1.13+.
func InternalErrorWith(err error) Error {
  twerr := NewError(Internal, err.Error())
  twerr = twerr.WithMeta("cause", fmt.Sprintf("%T", err)) // to easily tell apart wrapped internal errors from explicit ones
  return WrapError(twerr, err)
}
```

The way to use is by making the Twirp error first, and then wrapping the original:
```go
if err != nil {
  twerr := twirp.NewError(twirp.Malformed, "msg that is visible on the public")
  return nil, twirp.WrapError(twerr, err)
}
```

marioizquierdo:
Note: unwraping with Go 1.13+ errors package was implemented on #235

flimzy:
Your proposal would solve my problem. Would you like me to adjust this PR to follow that pattern?

marioizquierdo:
Let's wait until I can get another review from the team

marioizquierdo:
@flimzy I consulted and it looks good. Please, update the PR with the code that was proposed on the [previous comment](https://github.com/twitchtv/twirp/pull/229#issuecomment-633182638), that simply adds a new helper `WrapError(twerr Error, err error) Error` and refactors the existing helper `InternalErrorWith(err error) Error` with comments. And then it will be ready to merge and be released. Thank you so much for the suggestion!

flimzy:
@marioizquierdo I have rebased, and made the requested changes. Should I add some tests to cover the new functionality, as well?

marioizquierdo:
@flimzy Thanks for the tests. It looks good to merge!

PR #232, opened 2020-05-02
Go modules

alexplevako:
*Issue #, if available:*
#169 

*Description of changes:*
- This PR holds changes required for Go Modules support.
- Tools version locked in go.mod
- Installing protoc in _tools/


go.mod
```diff
@@ -0,0 +1,14 @@
+module github.com/twitchtv/twirp
```
marwan-at-work, line 1: this will not work because Twirp is tagged beyond v1. See https://github.com/twitchtv/twirp/issues/169 for an extended discussion

rochacon: Have not tested this, but wouldn't this work with the `+incompatible` tag in the `go.mod`?

```
module example.com/twirp-svc

require github.com/twitchtv/twirp v5.6.0+incompatible
```

marwan-at-work: It will not, because `+incompatible` indicates that the module does not have a go.mod file...once the repository has a go.mod file then you cannot opt-out of Semantic Import Versioning

albertorm95:
Will someone take a look at it?

Issue #233, opened 2020-05-03
Return response body when server returns invalid code


marwan-at-work:
When twirp cannot marshal an error into proper json, it conveniently includes the http response body as metadata as show here: https://github.com/twitchtv/twirp/blob/master/example/service.twirp.go#L591

Also this comments describes a good reason for why the body is included: https://github.com/twitchtv/twirp/blob/master/example/service.twirp.go#L543

However, sometimes an intermediary returns proper json, but improper twirp-json. In other words, the response can be something `{"msg": "gateway rate limit", "code": "rate-limit"}`. 

This will cause the Twirp client to successfully unmarshal the server returned json, but the json is not in the expected Twirp Error shape. Therefore this line of code gets executed: https://github.com/twitchtv/twirp/blob/master/example/service.twirp.go#L549:L552

Therefore, if Code is incorrect, we should still include the "body" in the json metadata so that it will not be swallowed. This way, our business logic can properly include the "body" into logging and observability tools to really know why the response was improper.

PR #234, opened 2020-05-03
Include response body in metadata when server returns invalid code

marwan-at-work:
*Issue #, if available:*
Fixes https://github.com/twitchtv/twirp/issues/233

*Description of changes:*

This adds the "body" metadata to the response when the err code is missing or invalid

I have updated `generator.go` and ran `retool do go generate ./...` -- hopefully this was the correct process. But definitely let me know if I need to do anything different. 

### DIFF
--- a/protoc-gen-twirp/generator.go
+++ b/protoc-gen-twirp/generator.go
@@ -538,7 +538,7 @@ func (t *twirp) generateUtils() {
 	t.P(`  errorCode := `, t.pkgs["twirp"], `.ErrorCode(tj.Code)`)
 	t.P(`  if !`, t.pkgs["twirp"], `.IsValidErrorCode(errorCode) {`)
 	t.P(`    msg := "invalid type returned from server error response: "+tj.Code`)
-	t.P(`    return `, t.pkgs["twirp"], `.InternalError(msg)`)
+	t.P(`    return `, t.pkgs["twirp"], `.InternalError(msg).WithMeta("body", string(respBodyBytes))`)
 	t.P(`  }`)
 	t.P(``)
 	t.P(`  twerr := `, t.pkgs["twirp"], `.NewError(errorCode, tj.Msg)`)
--- a/clientcompat/internal/clientcompat/clientcompat.twirp.go
+++ b/clientcompat/internal/clientcompat/clientcompat.twirp.go
@@ -722,7 +722,7 @@ func errorFromResponse(resp *http.Response) twirp.Error {
 	errorCode := twirp.ErrorCode(tj.Code)
 	if !twirp.IsValidErrorCode(errorCode) {
 		msg := "invalid type returned from server error response: " + tj.Code
-		return twirp.InternalError(msg)
+		return twirp.InternalError(msg).WithMeta("body", string(respBodyBytes))
 	}
 
 	twerr := twirp.NewError(errorCode, tj.Msg)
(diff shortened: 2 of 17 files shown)


protoc-gen-twirp/generator.go
```diff
@@ -538,7 +538,7 @@ func (t *twirp) generateUtils() {
 	t.P(`  errorCode := `, t.pkgs["twirp"], `.ErrorCode(tj.Code)`)
 	t.P(`  if !`, t.pkgs["twirp"], `.IsValidErrorCode(errorCode) {`)
 	t.P(`    msg := "invalid type returned from server error response: "+tj.Code`)
-	t.P(`    return `, t.pkgs["twirp"], `.InternalError(msg)`)
+	t.P(`    return `, t.pkgs["twirp"], `.InternalError(msg).WithMeta("body", string(respBodyBytes))`)
```
dpolansky, line 541: The error handling on this line doesn't distinguish between these two situations:
1.) The response body matches the shape of `twerrJSON`, but it has invalid data (such as an invalid error code)
2.) The response body does not match the shape of `twerrJSON` (maybe it's some other JSON response)

In the issue accompanying this PR, it seemed like 2) was relevant to your use case, but it seems like 1) was the original intent of this line. It may be a good idea to distinguish between those two cases, such that 2.) can reuse `twirpErrorFromIntermediary`. That helper function seems appropriate for that situation, and it includes additional metadata that would be useful for debugging.

Am I thinking about this correctly, @marwan-at-work?

marwan-at-work:
@dpolansky running `make` gave me a bunch of errors: 

```
github.com/twitchtv/twirp/internal/twirptest/no_package_name/no_package_name.proto: File not found.
no_package_name_importer.proto:3:1: Import "github.com/twitchtv/twirp/internal/twirptest/no_package_name/no_package_name.proto" was not found or had errors.
no_package_name_importer.proto:11:14: "Msg" is not defined.
no_package_name_importer.proto:11:28: "Msg" is not defined.
internal/twirptest/no_package_name_importer/gen.go:16: running "../../protoc_gen.sh": exit status 1
/src: warning: directory does not exist.
/src: warning: directory does not exist.
/src: warning: directory does not exist.
/src: warning: directory does not exist.
/src: warning: directory does not exist.
/src: warning: directory does not exist.
/src: warning: directory does not exist.
retool: fatal err: failed on 'go generate ./...': exit status 1
make: *** [generate] Error 1
```

I can just git-revert the changes under `_tools` or I can just only do the one change in `generator.go` and have you push an update to the branch? Whatever works 👍

protoc-gen-twirp/generator.go
marwan-at-work: @dpolansky actually my case is the one in `1` you mentioned above :) 
For case 2, we already do that because the json decoder does `DisallowUnknownFields` and also checks for `Code == ''` and `twirpErrorFromIntermediary` is indeed called. 

Would you still that function to be called? In other words: 

```
	errorCode := twirp.ErrorCode(tj.Code)
	if !twirp.IsValidErrorCode(errorCode) {
		msg := "invalid type returned from server error response: " + tj.Code
		return twirpErrorFromIntermediary(statusCode, msg, string(respBodyBytes))
	}
```

Let me know and I'm happy to do it :v:

dpolansky: Thanks for clarifying! I think this is fine as is for now.

marwan-at-work:
@dpolansky I reverted the `_tools` changes and re-pushed. Let me know if all is good. Thanks for the review!

PR #235, opened 2020-05-07
Implement Unwrap to allow go 1.13+ error handling

marioizquierdo:
This will fix issue #228

The `wrappedErr` type returned by the function `twirp.InternalErrorWith(err)`, that is used by twirp autogenerated clients, already implement the `Cause` method which works to unwrap internal errors with `github.com/pkg/errors`, but after Go 1.13, the standard way to handle wrapped errors is with the standard library: https://blog.golang.org/go1.13-errors

Implementing `Unwrap` will allow to check the original error with `errors.Is` and `errors.As`.


rhysh:
Thanks Mario!

As far as tests go: It should work with the Go 1.13+ core `errors` package; you might want a new test file with `// +build go1.13` at the top. (If we don't already have a test dependency on the github.com/pkg/errors package, it's ok to skip testing against that one.)

If a user wraps a `errors.New` value twice, I'd expect that `errors.Unwrap` takes off one layer at a time, returning the original after two calls.

I'd expect that calling `errors.Is` with the error the client returns from a canceled call returns true for `context.Canceled`. (Is that a feature we need to discuss separately? Are the changes to support it very involved?)

Does that sound like the right scope to you?

aconverse:
I think we are also going to need to add `Unwrap()` to the `wrappedError` type in the generated code as well, if we want the standard library errors package to be able to "see" through it and match errors with `errors.Is()`/`As()`. 

e.g. https://github.com/twitchtv/twirp/blob/v5.10.1/example/service.twirp.go#L606
https://github.com/twitchtv/twirp/blob/v5.10.1/protoc-gen-twirp/generator.go#L598

marioizquierdo:
@rhysh I'll try to make a test with an autogenerated client using context cancelled, but if it gets too complicated I also think it is ok to leave the test out; we don't have a dependency on github.com/pkg/errors, and the specification to implement the Unwrarp method is clear enough.

@aconverse good catch! right. That kind of sucks because it means services will need to update the clients in order to take advantage of `errors.Is/As`. But that's the only way we can move forward. I'll update the client generator, and add the "need updated clients" caveat in the release notes.

rhysh:
> we don't have a dependency on github.com/pkg/errors, and the specification to implement the Unwrarp method is clear enough.

Right, don't add a dependency on github.com/pkg/errors.

But please do test against Go 1.13+'s errors package. It's OK to depend on that in a file that's marked with `// +build go1.13`.

marioizquierdo:
Done.
 * Implement Unwrap on generated code for clients `wrappedError`
 * Implement Unwrap on generated code for service panics resulting in internal errors
 * Implement Unwrap on internal error returned by `twirp.InternalErrorWith`
 * Add go1.13 test to unwrap twirp.InternalErrorWith
 * Add go1.13 test for context.Canceled error, using errors.Is to check
 * Add go1.13 test for wrapped client tests using errors.Is to check

marioizquierdo:
@rhysh, @aconverse did you have a change to take a look? this branch should be ready to merge

aconverse:
Looks good to me

I tired this in my project and it lets me remove a bunch of extra unwrapping code and just use errors.Is()

PR #249, opened 2020-07-08
Use protojson instead of jsonpb

cyx:
As noted in their docs:

> Deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead.

See: https://godoc.org/github.com/golang/protobuf/jsonpb

This PR switches the JSON marshaling strategy to instead use protojson, which is the new blessed way for managing the JSON serialization strategy for protobufs.


cyx:
This actually doesn't work as simply as I assumed -- because protojson is using v2 and twirp is completely on v1. Are there other efforts to switch to v2?

dpolansky:
Hi @cyx, Twirp is indeed generating code using v1 today. There are challenges with changing that with respect to backwards compatibility. If you have interest in seeing v2 in Twirp, could you please make an issue and provide context on the value you'd get from it?

cyx:
Yeah I'm hesitating myself. the primary value is that it's v1 but given that Google said they would support that version indefinitely the only downside is really having both dependencies in the application code.

dpolansky:
Closing this for now, this should be discussed further in an issue.

PR #251, opened 2020-07-15
fix context is used in response hook after request

### LINKED ISSUE #250: In client hooks context from RequestPrepared is not received in ResponseReceived
As `callClientRequestPrepared` is called within `doProtobufRequest` and that function does not return the context the context provided to `ResponseReceived` "only" has the values from the generated RPC client call and everything prior to that but nothing from inside `doProtobufRequest`.

mspindelhirn:
*Issue #, if available:*
Fixes #250

*Description of changes:*
Context object is returned from doJSONRequest and doProtobufRequest so changes made to context are provided to ResponseReceived client hook


protoc-gen-twirp/generator.go
```diff
@@ -711,29 +711,30 @@ func (t *twirp) generateUtils() {
 	t.P()
 
 	t.P(`// doProtobufRequest makes a Protobuf request to the remote Twirp service.`)
-	t.P(`func doProtobufRequest(ctx `, t.pkgs["context"], `.Context, client HTTPClient, hooks *`, t.pkgs["twirp"], `.ClientHooks, url string, in, out `, t.pkgs["proto"], `.Message) (err error) {`)
+	t.P(`func doProtobufRequest(ctx `, t.pkgs["context"], `.Context, client HTTPClient, hooks *`, t.pkgs["twirp"], `.ClientHooks, url string, in, out `, t.pkgs["proto"], `.Message) (returnCtx `, t.pkgs["context"], `.Context, err error) {`)
```
dpolansky, line 714: I'm not sure why these return variables were named to begin with, since the code doesn't use a "naked return". Could we just make these unnamed returns instead and declare the error variable at the top?

mspindelhirn: There is one defer function that sets `err`:
https://github.com/twitchtv/twirp/pull/251/files/b1c5f0de3fd57b7de3e15eda135698b4cfbeaf08#diff-a3a2b0741a88a528d33eb1f40bef6b04L742
```
t.P(`  defer func() {`)
t.P(`    cerr := resp.Body.Close()`)
t.P(`    if err == nil && cerr != nil {`)
t.P(`      err = wrapInternal(cerr, "failed to close response body")`)
t.P(`    }`)
t.P(`  }()`)
```

That is right now the only reason to have named returns. 
I could move the first line `returnCtx = ctx` into the defer. Unsure what is best.
I could also switch to a complete naked return style by changing all current error returns.

dpolansky: Your test case looks great, thanks for adding that! Also I totally missed this defer block earlier, so thanks for pointing it out.

What do you think of just making the return signature `(_ context.Context, err error)`? This seems slightly more clear, since we're only relying on the named behavior for one of the values.

mspindelhirn: Changed the signature and rebased the change onto the main PR commit.

Also fixed a minor thing in the test. retool requires handling of all error returns and I ignored the service test call result so I added a check for that so that retool is happy.

mspindelhirn:
All squashed up and rebased against master again .. Reordered and squashed some prior commits to seperate tests and implementation for review and that included the other commit by accident ( as rebaseing root point ) 

All fixed up now with a rebase against latest master. Thanks for the review :+1:

Issue #244, opened 2020-06-18
Spec doesn't mention CamelCasing the URL


ofpiyush:
The way URLs are defined makes one believe the exact name of service is accepted. https://github.com/twitchtv/twirp/blob/03ac4c/docs/spec_v5.md#urls

In practice however both service and method names are changed. https://github.com/twitchtv/twirp/blob/03ac4c/protoc-gen-twirp/generator.go#L956

Should I add this to the spec for clarity and such that others do not get blind-sided by it while implementing?

@khevse discovered it in [twirpy](https://github.com/verloop/twirpy/pull/18)
I believe at least the [rust](https://github.com/cretz/prost-twirp/blob/master/src/service_gen.rs#L69), [swagger](https://github.com/elliots/protoc-gen-twirp_swagger/blob/master/genswagger/template.go#L673), [TypeScript](https://github.com/larrymyers/protoc-gen-twirp_typescript/blob/master/generator/minimal/client.go#L266), [Dart](https://github.com/apptreesoftware/protoc-gen-twirp_dart/blob/master/generator/client.go#L348) and [php](https://github.com/twirphp/twirp/blob/d556452006c63786e19ae2c8f16acd255f7292dc/protoc-gen-twirp_php/internal/gen/generator.go#L107) implementations suffer from this. 

Aside: The method itself is marked internal API. It might be easier for implementations if it were public https://github.com/twitchtv/twirp/blob/03ac4c/internal/gen/stringutils/stringutils.go#L71

ofpiyush:
Now I am more confused.

For the proto

```protobuf
package simple
service hello {
    rpc world(msg) returns(msg);
}
```

Is the desired behaviour:
1. Use service and method names as is. `/twirp/simple.hello/world`
2. CamelCase service name but not method name. `/twirp/simple.Hello/world`
3. CamelCase both. `/twirp/simple.Hello/World`

spenczar:
I think the spec points to 1. Use service and method names as is. I think protoc-gen-twirp is in the wrong. If it switches, though, it might cause compatibility problems; accepting both (again, just in protoc-gen-twirp) would soften the blow.

spenczar:
To be crystal clear: I no longer work at Twitch, and I'm not in charge around here. I haven't even had a chance to write much Go in 6 months. My opinion is just that, my opinion.

[...]

PR #257, opened 2020-07-30
Add support for spec compatible paths in generated server code

ofpiyush:
*Issue #, if available:*
#244 
*Description of changes:*
* Adds support for spec compatible paths in generated server code.
* Updates protoc to 3.12.4

No backwards incompatible changes have been made. Specifically:
* Go generated clients haven't been changed and will continue to send requests to the same URL as before.
* Go generated server will continue to serve on the CamelCased URLs it did before.
* `<ServiceName>PathPrefix` constant is unchanged as of now.
* `svc.PathPrefix()` still returns the CamelCased value.

Caveat: If someone is using a mux, they'll need to add the literal cased URL for handling manually as the intent was to not add anything to the API surface. [For context: start reading from here](https://github.com/twitchtv/twirp/pull/257#issuecomment-675220875)


protoc-gen-twirp/generator.go
```diff
@@ -1067,6 +1067,13 @@ func pathPrefix(file *descriptor.FileDescriptorProto, service *descriptor.Servic
 // pathFor returns the complete path for requests to a particular method on a
 // particular service.
 func pathFor(file *descriptor.FileDescriptorProto, service *descriptor.ServiceDescriptorProto, method *descriptor.MethodDescriptorProto) string {
+	return pathPrefix(file, service) + method.GetName()
+}
+
+// pathForCamelCompat returns the old CamelCase complete path for requests to a
+// particular method on a particular service.
+// This method is here to ensure compatibility with older clients.
```
dpolansky, line 1075: This comment isn't quite accurate, since clients will still make requests with camel-cased paths. This ultimately should link to documentation around this issue (which doesn't exist just yet).

I need to poke some folks internally about the proper place for these docs, and then we can update this comment to direct users to it.

protoc-gen-twirp/generator.go
```diff
@@ -1104,8 +1111,13 @@ func (t *twirp) generateServerRouting(servStruct string, file *descriptor.FileDe
 	t.P(`  switch req.URL.Path {`)
 	for _, method := range service.Method {
 		path := pathFor(file, service, method)
+		pathCamel := pathForCamelCompat(file, service, method)
 		methName := "serve" + stringutils.CamelCase(method.GetName())
-		t.P(`  case `, strconv.Quote(path), `:`)
+		caseCondition := strconv.Quote(path)
```
dpolansky, line 1116: Could you add a test in `internal/twirptest` that verifies this new behavior? You will also need to run `make` to re-generate the code with these changes

ofpiyush:
`protobuf@3.11.0` is not available on brew anymore.

`protobuf@3.12.4` didn't seem to break anything so I've updated that in this PR as well. 

stdin stdout code in `pycompat` broke with python 3.8 on my local so I've updated it.

Let me know if I should force install 3.11.0 or change pycompat in any way.

internal/twirptest/snake_case_names/compatibility_test.go
```diff
@@ -0,0 +1,68 @@
...
+// located at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package snake_case_names
```
dpolansky, line 14: I think we should just create a separate test for this instead of reusing this existing test scenario for clarity. In addition, it would allow us to test the literal names for both the service and method names. As it stands, `TestCompatibilityWithSpecCompatibleClient` only tests the literal method name and not the service name.

protoc-gen-twirp/generator.go
dpolansky: I made #262, which will add a page to the docs about this known issue.

internal/twirptest/snake_case_names/compatibility_test.go
ofpiyush: I assume this comment means a green signal for adding the new method `RegisterHaberdasherServer`.

Awesome!

dpolansky: What is this referring to? I don't know what `RegisterHaberdasherServer` means.

ofpiyush: First half of this comment. https://github.com/twitchtv/twirp/issues/244#issuecomment-666046457

While you're at it, please also go through the first half of this comment as it can completely resolve the issue.

https://github.com/twitchtv/twirp/issues/244#issuecomment-673935091

ofpiyush:
I've moved test to it's own package. Waiting for word on changes for service name and client hook.

I'll copy paste relevant parts from the comments to have context here.

## Server

This will need work from the user as well in cases where they're using a mux. We can manage method names without any changes from the user, but service name will need changes from the user.

One possible path which will contain this to 1 user change only, is to add a new method:

```go
example.RegisterHaberdasherServer(mux, handler)
```

Then we can keep changing code inside that function over releases and eventually into the new major version(if that ever happens) while making sure backwards compatibility is maintained.

## Client

The next best option I have is to generate a compatibility client hook and use a map to override URL path like I did on the test [here](https://github.com/ofpiyush/twirp/blob/312779fea1d5e901dd7bf906032eef1c78b8a570/internal/twirptest/snake_case_names/compatibility_test.go#L52-L57).

User experience could look something like:
```go
client := example.NewHaberdasherProtobufClient(url, client, twirp.WithClientHooks(example.CompatibilityHook))
```

internal/twirptest/snake_case_names/compatibility_test.go
ofpiyush: Resolved this to move conversation to main thread.

dpolansky:
Thanks for the refresher @ofpiyush! I read those comments, but didn't quite understand what you meant with regard to the server-side aspect until now. I think we should proceed with just addressing the method names for now, as addressing the service names will be complex and expands the API surface-area, which we try to be cautious about. Does that sound reasonable to you?

With regard to the compatibility hook, I would be hesitant to recommend it as a general solution to this problem because although it gets the job done, the hooks aren't intended for this purpose and it could lead to confusion. Hooks are generally not intended to mutate requests or responses that significantly. For now, I'd feel better about telling users to change their protobuf definitions if this issue affects them than tell them to use hooks to solve the problem. My opinion is that if this issue is painful enough for many Twirp users, we would then release a new major version with either changes to the spec or changes to the Go implementation.

---

Thanks for moving the test case to a separate package. Once we publish the docs PR, could we just update that comment to direct users there as we discussed previously? Thanks again for addressing the feedback.

ofpiyush:
> I think we should proceed with just addressing the method names for now

We could do slightly better and add code to handle it in non-mux cases and we could drop a line of gotcha of documentation for people who'd use a mux.

The next best option to a new method is to add a new PathPrefix constant.

Mux user experience would look something like:
```go
mux.Handle(example.HaberdasherPathPrefix, twirpHandler)
mux.Handle(example.HaberdasherSomeNamingConventionPathPrefix, twirpHandler)
```

I see the `RegisterHaberDasherServer` as a better alternative because that is a one time change for the user. It would open up path for other changes as well like #263 and eventually a smoother transition to v6.

> because although it gets the job done, the hooks aren't intended for this purpose and it could lead to confusion

Then I should change my test as well. What alternative would you suggest for the test?

The next best option I have is to do it as a http client wrapper. User experience would look something like:

```go
client := example.NewHaberdasherProtobufClient(url, example.HaberdasherCompatibilityClient(client))
```
---------------------------------------------------------------------------------
>  For now, I'd feel better about telling users to change their protobuf definitions

I think the fundamental assumption you're working with is that these poorly named services (even in go itself) are not in production or given out to third party.

I come from a world where some APIs are already public, our python servers were live with other python clients before go clients were introduced. We weren't even explicitly following protobuf style guide. We just happened to be fortunate that the random convention I followed isn't incompatible.

I think making that requirement part of the spec would be a much better alternative if adding small amounts of code to the API surface is not acceptable.

dpolansky:
@ofpiyush I'd like to focus on just making this incremental improvement to handle the method names for now.

> Then I should change my test as well. What alternative would you suggest for the test?
> 
> The next best option I have is to do it as a http client wrapper. User experience would look something like:
> 
> ```go
> client := example.NewHaberdasherProtobufClient(url, example.HaberdasherCompatibilityClient(client))
> ```

I think this is preferable to modifying the request through hooks, so the test could use this method. I'd like to avoid adding additional scope to this PR beyond that though, such as providing the HTTP client implementation in the generated code as you demonstrated here.

internal/twirptest/spec_compatibility/compatibility_test.go
```diff
@@ -0,0 +1,67 @@
...
+// located at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package spec_compatibility
```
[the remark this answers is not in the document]
ofpiyush: That was where it was! Check the comments above, haha!

@dpolansky asked me to make it into another package.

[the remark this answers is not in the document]
marioizquierdo: Oh, sorry for the confusion. Yes I read that comment but I think it was based on previous context (the PR had a larger scope before). If it is still not a good idea (or I didn't understand the comment well), then just rename this test. No big deal :P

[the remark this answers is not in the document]
ofpiyush: The general vibe I got was all the future tests that will be necessary to ensure spec compatibility will go into that package as well.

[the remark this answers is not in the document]
ofpiyush: I've moved it back.

protoc-gen-twirp/generator.go
```diff
@@ -1060,14 +1060,33 @@ func (t *twirp) generateServer(file *descriptor.FileDescriptorProto, service *de
...
+	return pathPrefix(file, service.GetName()) + method.GetName()
+}
+
+// pathForGoClient returns the CamelCase complete path for requests to a
+// particular method on a particular service.
+// This method is here to ensure compatibility with the go client.
+
+func pathForGoClient(file *descriptor.FileDescriptorProto, service *descriptor.ServiceDescriptorProto, method *descriptor.MethodDescriptorProto) string {
+	return pathPrefix(file, stringutils.CamelCase(service.GetName())) + stringutils.CamelCase(method.GetName())
+}
```
marioizquierdo, line 1080: ^ forgot to remove this one

protoc-gen-twirp/generator.go
```diff
@@ -1078,7 +1097,7 @@ func (t *twirp) generateServerRouting(servStruct string, file *descriptor.FileDe
 	t.P(`// `, pathPrefixConst, ` is used for all URL paths on a twirp `, servName, ` server.`)
 	t.P(`// Requests are always: POST `, pathPrefixConst, `/method`)
 	t.P(`// It can be used in an HTTP mux to route twirp requests along with non-twirp requests on other routes.`)
-	t.P(`const `, pathPrefixConst, ` = `, strconv.Quote(pathPrefix(file, service)))
+	t.P(`const `, pathPrefixConst, ` = `, strconv.Quote(pathPrefix(file, stringutils.CamelCase(service.GetName()))))
```
marioizquierdo, line 1100: This is the same as using `servName`, but it's fine, being explicit may be more readable.

ofpiyush: Yes I've been trying to keep strings and go variables separate so that later one can change these easier if necessary.

protoc-gen-twirp/generator.go
ofpiyush: aah yes! removed it now

internal/twirptest/spec_compatibility/compatibility_test.go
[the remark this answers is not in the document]
marioizquierdo: > The general vibe I got was all the future tests that will be necessary to ensure spec compatibility will go into that package as well.

I see. But I would argue that is not a good way to group tests (this may be just my opinion). You may end up having only one specific test (non-camel-cased-names) under a generic umbrella term (spec-compat). Or you may end up having too many unrelated tests under the same umbrella term. We just don't know, so it is usually best not to try guessing the future too much, and just name things for what they are. If things change too much in the future, well, then we should refactor, but it is better to refactor later, at the right time, when you know what works better. A good rule of thumb to know when it's time to refactor is the [rule of three](https://blog.codinghorror.com/rule-of-three/) 🤓

marioizquierdo:
This PR will be part of the upcoming v7 release.
@ofpiyush are you happy with all the changes? I can merge this in the release branch and get it ready with all other changes.

ofpiyush:
Yes. Feel free to merge it.


V7?

marioizquierdo:
Yes, we are going to skip v6 because there was a previous [version draft](https://twitchtv.github.io/twirp/docs/spec_v6.html) that was never fully implemented.

ofpiyush:
This PR can safely be added to V5. Can we add client support in V7 as well?

marioizquierdo:
Yes, this is backwards compatible. And the client support needs to be backwards compatible as well (through a client option, with default value that works like before). I'm including this on next release for simplicity (I'll spend some time this week managing the release). If you want, please submit another PR with the client side of this so we can bundle it together with the v7 release 🙏  😺 . If you don't have time right now, it's fine, we can handle that part later as well.

PR #268, opened 2020-09-08
Update context from callError

### LINKED ISSUE #267: ServerHooks Context bug: Context from Error is not passed to ResponseSent
There seems to be a bug in the go code generation that leads to context updates in `Error` server hooks failing to reach the `ResponseSent` hook.

https://github.com/twitchtv/twirp/blob/987a500c277f0334c8fed1d4986b4d4d874ab649/example/service.twirp.go#L273-L278

`callError` on line 276 returns a new context, which is ignored, rather than overriding `ctx`.

~~This behavior is critical for finishing spans correctly in the [twirp/datadog integration](https://github.com/DataDog/dd-trace-go/blob/6d7c52cd8ab1ea309fd54c1a2ef472fadb84429c/contrib/twitchtv/twirp/twirp.go#L204-L205). In general, this behavior is critical any time the error hook wants to pass state down to the response sent callback.~~

Edit: This bug only manifests in a rare edge case, when writing the response fails. Normal application errors go through `writeError` in the generated code, which does update `ctx` correctly based on `callError` before calling `callResponseSent`.

mickeyreiss:
Assuming #267 is an valid bug report, this PR provides a fix to `protoc-gen-twirp`, so that the context returned by `Error` server hooks (on write errors) is subsequently passed into `ResponseSent`.

Fixes #267.


marioizquierdo:
Looks good to me. We should include this on the next release, thanks!

mickeyreiss:
@marioizquierdo Is there anything left to do on my end, in order to get this landed?

marioizquierdo:
Nah, you are good. I will help get this merged sometime this week.

PR #270, opened 2020-09-14
ResourceExhausted error code to HTTP 429 status code

marioizquierdo:
Followup after PR: #258

After all the analysis done in #258, we decided to re-map `ResourceExhausted` to 429 `Too Many Requests`, instead of adding a new status code. See PR comments for more details.

### Backwards compatibility

This is a non-backwards compatible change, although it may only affect servers and their middleware; clients are not affected by the HTTP status code. Note that, since the update happens on the `twirp` package, the update takes effect when the package is updated.

Backwards compatibility notes about changing the HTTP mapping code of resource_exhausted from `403` to `429`:

 * ✅  : New clients can handle all codes from old services (HTTP codes are not used to deserialize errors).
 * ⚠️  : Old clients in Go handle 429 errors from intermediaries (non-twirp services or middleware) as `unavailable` ([see code](https://github.com/twitchtv/twirp/blob/d0e66eecce5a8ef83db96ad71802978942598b0a/example/service.twirp.go#L578)). New clients will handle them as `resource_exhausted`.
 * ⚠️  : Old clients in different languages may have different ways to handle errors from intermediaries. We should file an issue on other implementations to warn them of the update.
 * ⚠️  : Old clients receiving errors from new services will still map `resource_exhausted` as 403, even if in the server it is mapped to 429 (the mapping is done in code).
 * ❌  : Middleware and proxies that depend on HTTP codes need to be reviewed and updated to handle 429. This update is not easy, and in some cases may be impossible (maybe the middleware is managed by a different team). During the update, the middleware needs to handle both 429 and 403 during codes properly.
 * ❌  : Monitoring based on HTTP codes needs to be reviewed.

(✅  = it just works, ⚠️  = corner cases may need attention, ❌  = code depending on previous behavior needs to be updated)


docs/spec_v7.md
```diff
@@ -208,7 +208,7 @@ corresponding HTTP Status Code for the response.
 | already_exists      | 409 | An attempt to create an entity failed because one already exists.
 | permission_denied   | 403 | The caller does not have permission to execute the specified operation. It must not be used if the caller cannot be identified (use "unauthenticated" instead).
 | unauthenticated     | 401 | The request does not have valid authentication credentials for the operation.
-| resource_exhausted  | 403 | Some resource has been exhausted, perhaps a per-user quota, or perhaps the entire file system is out of space.
+| resource_exhausted  | 429 | Some resource has been exhausted, perhaps a per-user quota, or perhaps the entire file system is out of space.
```
thesilentg, line 211: Is it worth mentioning rate-limiting here?

protoc-gen-twirp/generator.go
```diff
@@ -611,7 +611,9 @@ func (t *twirp) generateUtils() {
 	t.P(`      code = `, t.pkgs["twirp"], `.PermissionDenied`)
 	t.P(`    case 404: // Not Found`)
 	t.P(`      code = `, t.pkgs["twirp"], `.BadRoute`)
-	t.P(`    case 429, 502, 503, 504: // Too Many Requests, Bad Gateway, Service Unavailable, Gateway Timeout`)
+	t.P(`    case 429: // Too Many Requests`)
+	t.P(`      code = `, t.pkgs["twirp"], `.ResourceExhausted`)
+	t.P(`    case 502, 503, 504: // Too Many Requests, Bad Gateway, Service Unavailable, Gateway Timeout`)
```
thesilentg, line 616: Remove "Too Many Requests" from this line

thesilentg (changes requested):

docs/spec_v7.md
marioizquierdo: What about adding "exhausted or rate-limited"?
```
Some resource has been exhausted or rate-limited, perhaps a per-user quota, or perhaps the entire file system is out of space
```

thesilentg (approved):

PR #274, opened 2020-09-15
Add version_constant and check from generated code

marioizquierdo:
The PR #264 includes new server options in the `twirp` package that are used by the generated code. If the `protoc-gen-twirp` plugin used to generate code uses the latest version, but the `twirp` package in the project is an older version, the generated code will fail with a compile-time error. Without this constant, the error is something like `clientOpts.PathPrefix() function does not exist`. With this constant, the error is more descriptive, like `twirp.TwirpPackageIsVersion7 does not exist`, which better points to the solution.

This pattern is similar to protobuf generated code, that uses compile-time assertions like `proto.ProtoPackageIsVersion4` to make sure the right version of the proto library is included in the destination project.

### DIFF
--- a/protoc-gen-twirp/generator.go
+++ b/protoc-gen-twirp/generator.go
@@ -223,6 +223,8 @@ func (t *twirp) generate(file *descriptor.FileDescriptorProto) *plugin.CodeGener
 		t.generateUtilImports()
 	}
 
+	t.generateVersionCheck(file)
+
 	// For each service, generate client stubs and server
 	for i, service := range file.Service {
 		t.generateService(file, service, i)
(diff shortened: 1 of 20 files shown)


protoc-gen-twirp/generator.go
```diff
@@ -243,6 +245,14 @@ func (t *twirp) generate(file *descriptor.FileDescriptorProto) *plugin.CodeGener
 	return resp
 }
 
+func (t *twirp) generateVersionCheck(file *descriptor.FileDescriptorProto) {
+	t.P(`// This is a compile-time assertion to ensure that this generated file`)
+	t.P(`// is compatible with the twirp package used in your project.`)
+	t.P(`// A compilation error at this line likely means your copy of the`)
+	t.P(`// twirp package needs to be updated.`)
+	t.P(`const _ = `, t.pkgs["twirp"], `.TwirpPackageIsVersion7`)
```
DennisSSDev, line 253: This seems very similar to the protobuf assertion line, so I think this is a good idea to add so we can also validate the twirp version. LGTM!

DennisSSDev (approved):

Issue #265, opened 2020-09-01
Proposal: Request/Response interceptors


amckinney:
In gRPC, interceptors are able to access the request and response objects from
the client and/or server.

Twirp supports similar middleware functionality via its `ClientHooks` and
`ServerHooks` constructs, but it's not currently possible to inspect the
request or response objects like gRPC.

This proposes a new `Interceptors` option for both the `ClientHooks` and
`ServerHooks` types.

For `ClientHooks` the user-provided interceptors would be able to take action
 * After the request structure is marshaled
 * Before the request is sent over the wire
 * After the response is returned

 Borrowing from Twirp's generated code, in `doProtobufRequest`,
 this would look something along the lines of the following:

 ```go
 func doProtoRequest(...) (context.Context, error) {
  req, err := newRequest(ctx, url, reqBody, "application/protobuf")
  if err != nil {
    return ctx, wrapInternal(err, "could not build request")
  }
  ctx, err = callClientRequestPrepared(ctx, hooks, req)
  if err != nil {
    return ctx, err
  }

  req = req.WithContext(ctx)
  method := twirp.ChainInterceptors(hooks.Interceptors...)(doProtobufMethod(out))
  if _, err := method(ctx, req); err != nil {
    return ctx, wrapInternal(err, "oops")
  }

  return ctx, nil
}

// doProtobufMethod returns an twirp.Method that performs the client call.
// This method should always occur last in the interceptor chain.
func doProtobufMethod(out proto.Message) twirp.Method {
  return func(ctx context.Context, req interface{}) (interface{}, error) {
    resp, err := client.Do(req)
    if err != nil {
      return ctx, wrapInternal(err, "failed to do request")
    }

    defer func() {
      cerr := resp.Body.Close()
      if err == nil && cerr != nil {
        err = wrapInternal(cerr, "failed to close response body")
      }
    }()

    if err = ctx.Err(); err != nil {
      return ctx, wrapInternal(err, "aborted because context was done")
    }

    if resp.StatusCode != 200 {
      return ctx, errorFromResponse(resp)
    }

    respBodyBytes, err := ioutil.ReadAll(resp.Body)
    if err != nil {
      return ctx, wrapInternal(err, "failed to read response body")
    }
    if err = ctx.Err(); err != nil {
      return ctx, wrapInternal(err, "aborted because context was done")
    }

    if err = proto.Unmarshal(respBodyBytes, out); err != nil {
[...]

PR #276, opened 2020-09-17
Add interceptors

bufdev:
Implements #265


Makefile
```diff
@@ -28,7 +28,7 @@ test_python_client: generate build/clientcompat build/pycompat
 
 setup:
 	./install_proto.bash
-	GOPATH=$(CURDIR)/_tools go install github.com/twitchtv/retool/...
+	GO111MODULE=off GOPATH=$(CURDIR)/_tools GOBIN=$(CURDIR)/_tools/bin go get github.com/twitchtv/retool
```
bufdev, line 31: This is needed for any development system that has `GO111MODULE` and `GOBIN` set, and has no effect on other systems for Twirp's `Makefile` logic.

protoc-gen-twirp/generator.go
```diff
@@ -1335,6 +1347,32 @@ func (t *twirp) generateServerProtobufMethod(service *descriptor.ServiceDescript
...
+	t.P(`      )(ctx, req)`)
+	t.P(`      if resp != nil {`)
+	t.P(`        typedResp, ok := resp.(*`, outputType, `)`)
+	t.P(`        if !ok {`)
+	t.P(`          return nil, `, t.pkgs["twirp"], `.InternalError("could not convert to a *`, outputType, `")`)
+	t.P(`        }`)
+	t.P(`        return typedResp, err`)
+	t.P(`      }`)
+	t.P(`      return nil, err`)
+	t.P(`    }`)
```
marioizquierdo, line 1373: I have to say: if only Go had generics 😅

protoc-gen-twirp/generator.go
```diff
@@ -1335,6 +1347,32 @@ func (t *twirp) generateServerProtobufMethod(service *descriptor.ServiceDescript
...
+	t.P(`          if !ok {`)
+	t.P(`            return nil, `, t.pkgs["twirp"], `.InternalError("could not convert to a *`, inputType, `")`)
+	t.P(`          }`)
+	t.P(`          return s.`, servName, `.`, methName, `(ctx, typedReq)`)
+	t.P(`        },`)
+	t.P(`      )(ctx, req)`)
+	t.P(`      if resp != nil {`)
+	t.P(`        typedResp, ok := resp.(*`, outputType, `)`)
+	t.P(`        if !ok {`)
+	t.P(`          return nil, `, t.pkgs["twirp"], `.InternalError("could not convert to a *`, outputType, `")`)
```
marioizquierdo, line 1368: add more context to this error: `twirp.InternalError("failed type assertion resp.(*Msg) when calling interceptor handler")`. This error should never happen, but if it does, we better have a long descriptive error to make it as easy to find as possible (same for the type assertion on the request)

middleware_test.go
```diff
@@ -0,0 +1,71 @@
...
+			want:         "ax1",
+		},
+		{
+			interceptors: []Interceptor{interceptor1, interceptor2},
+			want:         "abx21",
+		},
+		{
+			interceptors: []Interceptor{interceptor1, interceptor2, interceptor3},
+			want:         "abcx321",
+		},
```
marioizquierdo, line 61: add another case using the same interceptor instance multiple times:
```go
{
  interceptors: []Interceptor{interceptor1, interceptor1, interceptor1},
  want:         "aaax111",
}
```

middleware.go
```diff
@@ -0,0 +1,44 @@
...
+// Method is a method.
+type Method func(ctx context.Context, request interface{}) (interface{}, error)
+
+// Interceptor is a interceptor.
+type Interceptor func(Method) Method
+
+// ChainInterceptors chains the Interceptors.
+//
+// Returns nil if interceptors is empty.
+func ChainInterceptors(interceptors ...Interceptor) Interceptor {
```
marioizquierdo, line 28: to be safe, we should remove nils from the array, which will be equivalent to ignoring nil interceptors in the Chain, which makes sense:
```go
intrs := make([]Interceptor, 0, len(interceptors))
for _, intr := range interceptors {
    if intr != nil {
        intrs = append(intrs, intr)
    }
}
```

Then also add a testcase to verify if nil interceptors are properly ignored.

middleware.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Method is a method.
+type Method func(ctx context.Context, request interface{}) (interface{}, error)
+
+// Interceptor is a interceptor.
+type Interceptor func(Method) Method
+
+// ChainInterceptors chains the Interceptors.
+//
+// Returns nil if interceptors is empty.
```
marioizquierdo, line 27: Let's remember to add more comments explaining `Method`, `Interceptor` and `ChainInterceptors`. The extended comments should contain some basic example. We should also add a new page for interceptors in the docs, and make sure it is linked from docs for [Hooks](https://twitchtv.github.io/twirp/docs/hooks.html), and from other pages that mention hooks.

middleware_test.go
```diff
@@ -0,0 +1,71 @@
...
+// located at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed on
+// an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package twirp
+
```
marioizquierdo, line 14: Let's settle on the word "interceptors". Rename these files to `interceptors.go` and `interceptors_test.go`

internal/twirptest/service_test.go
```diff
@@ -574,6 +575,48 @@ func TestErroringHooks(t *testing.T) {
 	})
 }
 
+func TestInterceptor(t *testing.T) {
+	interceptor := func(next twirp.Method) twirp.Method {
+		return func(ctx context.Context, request interface{}) (interface{}, error) {
```
marioizquierdo, line 580: verify that the context contains info about the request:

```go
methodName, _ := twirp.MethodName(ctx)
if methodName != "MakeHat" {
  return nil, fmt.Errorf("unexpected methodName: %q", methodName)
}
serviceName, _ := twirp.ServiceName(ctx)
if serviceName != "Haberdasher" {
  return nil, fmt.Errorf("unexpected serviceName: %q", serviceName)
}
packageName, _ := twirp.PackageName(ctx)
if packageName != "twirp.internal.twirptest" {
  return nil, fmt.Errorf("unexpected packageName: %q", packageName)
}
```

later, after calling the service, we should also verify that the context was annotated with the status code:

```go
response, err := next(ctx, request)

status, ok := twirp.StatusCode(ctx)
if status != "200" || !ok {
    return nil, fmt.Errorf("unexpected status in context: %q", status)
}
```

internal/twirptest/service_test.go
```diff
@@ -574,6 +575,48 @@ func TestErroringHooks(t *testing.T) {
...
+	if clientErr != nil {
+		t.Fatalf("client err=%q", clientErr)
+	}
+	if hat.Size != 3 {
+		t.Errorf("hat size expected=3 actual=%v", hat.Size)
+	}
+	if hat.Color != "bluexx" {
+		t.Errorf("hat color expected=bluexx actual=%v", hat.Color)
+	}
+}
```
marioizquierdo, line 618: Make sure to add test cases with no errors (using `&Size{Inches: 3}` on the `PickyHatmaker(3)` should return success with no errors. And also verify that calling with a different size (e.g. `&Size{Inches: 1}`) returns the proper error that can be casted to a `twirp.Error` type and the `twerr.Code()` is `twirp.InvalidArgument` (as returned by the PickyHatmaker fake implementation.

marioizquierdo (changes requested):
Nice. It is looking good 👍 
Looking forward to complete the implementation on the client side.

protoc-gen-twirp/generator.go
bufdev: Heh

protoc-gen-twirp/generator.go
bufdev: Done

middleware_test.go
bufdev: Done

middleware.go
bufdev: This isn't idiomatic Golang to me - I'd expect that users would make sure such inputs are not nil, and would want to be loudly broken if a nil `Interceptor` was put in here. However, I see that `ServerHooks` does this, so I buy the argument to stay consistent here. Updated.

middleware.go
bufdev: I added a bit more Godoc, and a skeleton for the `docs` folder - could you propose some more language you'd like to add here? Would be helpful.

middleware_test.go
bufdev: Done

internal/twirptest/service_test.go
bufdev: Added the first part, but the second part (status on context) isn't how contexts work - contexts are immutable, and we'd have to return a new context to do this (which isn't standard). I'd expect to get the status code off of the error, not the context, and all other middleware of this variety in the Golang ecosystem does the same.

internal/twirptest/service_test.go
bufdev: Done

bufdev:
Addressed comments, will start on client interceptors now.

bufdev:
Update: the client interceptors are now added, including testing.

marioizquierdo (approved):
Looks good to me. The initial documentation looks good, we can edit later.

marioizquierdo:
Will merge in master with the release v7.1 soon. Great job! 👍

internal/twirptest/service_test.go
kl09: good idea, thanks. Getting statusCode from context - doesn't work.
This is how I did:
```
twerr, ok := err.(twirp.Error)
if ok {
	statusCode = twirp.ServerHTTPStatusFromErrorCode(twerr.Code())
}
```

PR #277, opened 2020-09-19
Add client option for literal URLs

ofpiyush:
*Issue #, if available:*
#244 , #257

*Description of changes:*
Adds a new client option to use literal URLs instead of the default CamelCased ones.

The updated test in `internal/twirptest/snake_case_names` covers behaviour of both the client and the server.


ofpiyush:
@marioizquierdo I am not 100% sure on the name for the public method `WithClientLiteralCase`. If you have a better one, I'd be happy to change it.

Other than that, this PR is ready to merge

marioizquierdo:
I'm adding a patch to get it ready. We just conflicted merging from master at the same time heh. Will be ready in a few minutes and I'll merge. Thanks fro this work!

marioizquierdo:
Updated to keep the client implementation with a single list of `urls` (instead of having two lists), and when the Service and Method names are already CamelCased (like most examples in tests) keep the exact same generated code.

PR #285, opened 2020-11-04
Switch to go modules and initialize v8 of the project

gurudesu:
This sets up [go modules](https://blog.golang.org/v2-go-modules) for this repository and changes the base version to v8.

Additionally, this removes the checked in `vendor` folder of the project.

This resolves https://github.com/twitchtv/twirp/issues/169

## Changes to imports

Imports are now updated to `github.com/twitchtv/twirp/v8` for users of `v8.0.0` and above.

Other than that, no major changes for even users of `dep` as long as they are using a version of `go > LATEST - 3`.

Future major revisions will need to go into the `./v9` folder of the repository if there comes a need for another major revision.

## Disclaimer


gurudesu:
Opened this to start a conversation around our go modules approach.

Looks fairly heavy handed in terms of folks providing libraries that wrap twirp. They will have to update their libraries _every major version release_ which is not ideal. Exploring a few things so folks won't need to.

Issue #280, opened 2020-10-11
detecting import paths is incorrect for root vanity imports


marwan-at-work:
Perhaps the comment in this line of code describes the issue best: https://github.com/twitchtv/twirp/blob/ead44db746049716fda1702d896ebdc2db1e898b/protoc-gen-twirp/go_naming.go#L34

An import path may not actually have a slash in it. For example if you define a vanity import such as "go.mycompany.com" and then define your protobuf file at the root of the module as such: 

```protobuf
syntax = "proto3";
package mycompany;
option go_package = "go.mycompany.com;mycompany";

...
```

Twirp will end up generating the following Go package declaration: 

```go
package go_mycompany_com_mycompany

...
```

This is clearly incorrect. Twirp should instead see if there's a semi-colon in the go_package option and then return the left hand side as the import path and the right hand side as the package name.

amcohen-twitch:
Hi @marwan-at-work, thanks for bringing this to our attention. Our team was unaware of this expected behavior, and are having trouble finding documentation of it. By any chance, could you point us to any further details/docs?

marwan-at-work:
@amcohen-twitch if try to generate Go code with `--go_out=.` with the latest protoc-gen-go (v2) and notice the warning it will spit out: 

```
2020/11/22 16:42:07 WARNING: Missing 'go_package' option in "dbregistry.proto",
please specify it with the full Go package path as
a future release of protoc-gen-go will require this be specified.
See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.
```

So the docs it refers to is here: https://developers.google.com/protocol-buffers/docs/reference/go-generated#package

PR #288, opened 2020-11-22
Fix go_option that contains both path and pkg

marwan-at-work:
This PR adds a new check to the `go_option` parser to see if it includes a semi-colon. If it does, that means the user has already provided both the import path as well as the package name and that Twirp shouldn't try to deduce either of them. 

In protobuf-go's v2, this is going to be the new default. See the referenced issue for more details. 

Fixes https://github.com/twitchtv/twirp/issues/280

### DIFF
--- a/protoc-gen-twirp/go_naming.go
+++ b/protoc-gen-twirp/go_naming.go
@@ -31,6 +31,9 @@ func goPackageOption(f *descriptor.FileDescriptorProto) (impPath, pkg string, ok
 		return
 	}
 	ok = true
+	if bits := strings.Split(pkg, ";"); len(bits) == 2 {
+		return bits[0], bits[1], ok
+	}
 	// The presence of a slash implies there's an import path.
 	slash := strings.LastIndex(pkg, "/")
 	if slash < 0 {
--- a/protoc-gen-twirp/go_naming_test.go
+++ b/protoc-gen-twirp/go_naming_test.go
@@ -13,13 +13,15 @@
 
 package main
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+)
 
 func TestParseGoPackageOption(t *testing.T) {
 	testcase := func(in, wantImport, wantPkg string) func(*testing.T) {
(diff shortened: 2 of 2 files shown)


chetanvalagit:
@marwan-at-work can you please add some detail description as part of PR? 

Reference Example: https://github.com/twitchtv/twirp/pull/277

marwan-at-work:
@chetanvalagit done :)

chetanvalagit:
@marwan-at-work thank you very much. I am really sorry for the one more quick request: Can you please add last line too as shown in Reference Example: #277 if you are fine with that.

chetanvalagit:
@marwan-at-work thank you very much. I am really sorry for the one more quick request: Can you please add last line too as shown in Reference Example: #277 if you are fine with that

marwan-at-work:
@chetanvalagit done :v:

amcohen-twitch:
Upon review, we're generally on board with this implementation. Could you please add a test to the existing tests for this function? https://github.com/twitchtv/twirp/blob/424ec823543050237314da4efe3ac642abb618f6/protoc-gen-twirp/go_naming_test.go

marwan-at-work:
@amcohen-twitch I don't think the link you sent me actually points to the function I modified. The link you sent is testing the `parseGoPackageOption` function which actually seems to be doing the correct logic. While, my function was missing the same logic. 

Furthermore, that test function seems to be completely wrong, because it's shadowing all the test cases with empty strings. 

Would you like me to create a new test for my own change, and also fix the tests you pointed out above?

marwan-at-work:
@amcohen-twitch I fixed the linked test cases and added new ones for my particular fix. Happy to change things up if I missed something here.

advanderveer:
Just wanted to say that i'm looking forward to this change. Hopefully it can be merged soon! Thank you

marwan-at-work:
@amcohen-twitch any luck on the above? Thanks!

amcohen-twitch:
@marwan-at-work Apologies for the delay in followup, we were out of office due to the holidays. Upon subsequent review, we're good with merging this.

PR #289, opened 2020-12-04
Added some cleanup to docs and removed retool

DennisSSDev:
Ran my linter on the docs that were changed.

The linter is as follows:

Prettier 5.8.0 default settings.

Standouts:
- Changed the Go description to state that we support the latest 2 versions of Go
- removed mentions of `retool` since it's a deprecated package
- Some quick doc fixes on `brew` & `bashrc`
- Removed glide

Question - do we need to add an example of adding protoc-gen-go && protoc-gen-twirp via go modules? Since they are command line dependencies/code generators I don't think they'd be going into a go module...


marioizquierdo:
I'm cool with your linter, except for the bullet points updated from `*` to `-`. I mean, if it needs to be a dash then it needs to be, but I don't know why a linter would prefer that, everyone uses `*` right? (this may be a "tab vs spaces" war and I didn't know about it until now 🤣 )

DennisSSDev:
@marioizquierdo fixed 😃

marioizquierdo (commented):
Thanks!

PR #290, opened 2020-12-12
Add ErrorCode PayloadTooLarge

avinassh:
*Description of changes:*

Currently there is no way to send an error when the request body is too large. Added a new error `PayloadTooLarge`


marioizquierdo:
Adding a new error code is a non-backwards compatible change, and as such it needs a really good reason and great deal of support to be merged. The error codes are not meant to be exhaustive or map 1-1 with HTTP codes. If a payload is too large, the service could respond with code `InvalidArgument`. If more information is needed, use metadata.

For example:
```
return nil, twirp.NewError(twirp.InvalidArgument, "Payload too large").WithMeta("payload_limit_bytes", 999)
```

DennisSSDev:
@avinassh could you perhaps give a compelling example of why there should be an explicit error code for this case?

avinassh:
I thought Twirp maintains 1:1 mapping with HTTP codes. My bad, closing this issue 😅

Issue #293, opened 2021-02-23
Proposal: Check request context before returning internal error on server handler


shuangzzzz:
I have a client talking to a server with an ALB setup in the front. The client is causing the server to generate `unexpected EOF` error. The error was generated when the request body was read ([example link](https://github.com/twitchtv/twirp/blob/6d39013d250bb15fd7bd0d9e105340cb0ca1224a/example/service.twirp.go#L417)).

Golang will try to automatically upgrade the connection to use http2 if sending a https request. The interaction between client and server will be like the following: 
Client----request(http2)---->ALB----request(http1.1)---->Server
If context cancellation happens, client will send in a http2 RST_STREAM frame to ALB and ALB will close the connection to the server.

For http2, the request body might not be immediately sent after the header is written. With context cancellation, there is a chance that the request body is not sent or partially sent to the server. The server will generate the `unexpected EOF` error since the request body length does not match the `Content-Length` field in the header. 

This will result in an internal twirp error on the server side which will be interpreted as 500. This will drop the service availability and trigger alarms. To avoid this, the twirp server handler can check the context in the request before raising an internal error. In this case, `ctx.Err()` should return a context canceled error. Twirp should map this error to the code `Canceled`. If `ctx.Err()` does not return anything, I think we can return either `Malformed`.

PR #294, opened 2021-03-08
handling and returning valid error code for context.Canceled and cont…

chetanvalagit:
…ext.DeadlineExceeded error

*Issue #, if available: #293

*Description of changes:*
Proper error code return in case of context.Canceled (Error code will be returned twirp.Canceled) and context.DeadlineExceeded (Error code will be returned twirp. DeadlineExceeded), instead of 5xx 'Unexpected EOF'


internal/twirptest/proto/proto_test.go
```diff
@@ -13,8 +13,129 @@
...
+	"time"
+
+	twirp "github.com/twitchtv/twirp"
+)
 
 func TestCompilation(t *testing.T) {
 	// Test passes if this package compiles
 }
+
+type errReader int
```
marioizquierdo, line 33: Adding the test in `/proto/proto_test.go`,  feels a little arbitrary, because this is not related to testing an empty proto message. What about moving the test to `service_test.go`? all other arbitrary tests are there

internal/twirptest/proto/proto_test.go
```diff
@@ -13,8 +13,129 @@
...
+	ctx, _ = context.WithTimeout(ctx, 1*time.Millisecond)
+
+	// Make a request, that will endpoint method
+	req, _ := http.NewRequest(http.MethodPost, "http://testing:8080/twirp/twirp.internal.twirptest.proto.Svc/Send", errReader(0))
+	req.Header.Set("Accept", "application/protobuf")
+	req.Header.Set("Content-Type", "application/protobuf")
+	// Associate the cancellable context we just created to the request
+	req = req.WithContext(ctx)
+	// cancel context
+	time.Sleep(2 * time.Millisecond)
```
marioizquierdo, line 89: Instead of using time.Sleep, you should check the timeout channel like this:
```go
select {
case <-ctx.Done():
    // ctx.Err() will have context.DeadlineExceeded
case <-time.After(1 * time.Second):
    t.Fatalf("Waited too long for DeadlineExceeded ")
}
```

internal/twirptest/proto/proto_test.go
```diff
@@ -13,8 +13,129 @@
...
+	}
+
+	expectedErrMessage := "context deadline exceeded"
+	if !strings.Contains(string(respBytes), expectedErrMessage) {
+		t.Errorf("twirp client err has unexpected message %q, want %q", string(respBytes), expectedErrMessage)
+	}
+}
+
+func TestReadRequestError(t *testing.T) {
+	// Test passes if this package compiles
```
marioizquierdo, line 110: Should remove this comment. For the test name, what about something more explicit like `TestFailedToReadRequestBodyError` ?

marioizquierdo (approved):
LGTM.

Let's release this as `v7.2.0`

protoc-gen-twirp/generator.go
```diff
@@ -1334,6 +1334,13 @@ func (t *twirp) generateServerProtobufMethod(service *descriptor.ServiceDescript
 	t.P()
 	t.P(`  buf, err := `, t.pkgs["ioutil"], `.ReadAll(req.Body)`)
 	t.P(`  if err != nil {`)
+	t.P(`    if ctx.Err() == context.Canceled  {`)
+	t.P(`    		s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, "context canceled"))`)
```
rhysh, line 1338: Optional: It seems unnecessary to write out the meaning of the error three times here (checking equality, setting Twirp status, adding message), and another three times in the next block. What do you think about replacing the last one of those with a call to `context.Canceled.Error()`, either via `ctx.Err().Error()` (since we have just determined that the two are `==`), or via `if ctxErr := context.Canceled; ctx.Err() == cxtErr /* ... */ ctxErr.Error()`?

It's not critical to change, and I'm not thrilled with any of the options I've provided for changing it. But it seems like more repetition than we need.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,123 @@ func TestPanicFlushing(t *testing.T) {
...
+	}
+}
+
+func TestDeadlineExceededError(t *testing.T) {
+	// test case validate context deadline exceed error handling
+	var s Haberdasher
+	server := NewHaberdasherServer(s, nil)
+
+	ctx := context.Background()
+	ctx, _ = context.WithTimeout(ctx, 1*time.Millisecond)
```
rhysh, line 1580: Optional: When the ctx value is scoped to a function, WithTimeout and WithCancel usually get a `defer cancel()` right after to make that clear. For this, if it were my code, I'd give the CancelFunc a name (cancel) and immediately use it (as `defer cancel()`). Same for the previous test, though it already has a name there.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,123 @@ func TestPanicFlushing(t *testing.T) {
 		t.Errorf("twirp client err has unexpected message %q, want %q", twerr.Msg(), "Internal service panic")
 	}
 }
+
+type errReader int
```
rhysh, line 1531: Optional: It looks like the value of this never gets used (but looks in some places like it _might_). I'd make this a `struct{}` so it's clear that it holds no useful information.

And an extra aside: I'd suggest also https://golang.org/pkg/testing/iotest/#ErrReader , but I'm surprised to see that it's new in Go 1.16, so is still too new to use in Twirp (we need to support at least the previous Go version, and maybe the one before that too).

rhysh (changes requested):
The behavior change LGTM, but the tests are too far removed from how Twirp gets used in production. The tests should verify that we get the right behavior higher up in the stack, and demonstrate to readers (including our future selves) what that high-level behavior _is_.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,123 @@ func TestPanicFlushing(t *testing.T) {
...
+		t.Fatalf("Waited too long for DeadlineExceeded ")
+	}
+	w := httptest.NewRecorder()
+	server.ServeHTTP(w, req)
+
+	// validate error code
+	expectedErrCode := twirp.ServerHTTPStatusFromErrorCode(twirp.DeadlineExceeded)
+	if w.Code != expectedErrCode {
+		t.Errorf("twirp ErrorCode expected to be %q, but found %q", expectedErrCode, w.Code)
+	}
```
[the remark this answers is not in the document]
chetanvalagit: Checking as part of server hooks

protoc-gen-twirp/generator.go
chetanvalagit: Done

internal/twirptest/service_test.go
chetanvalagit: Agree. Done

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,123 @@ func TestPanicFlushing(t *testing.T) {
...
+	req = req.WithContext(ctx)
+
+	// wait for deadline exceed timeout in context object
+	select {
+	case <-ctx.Done():
+		// ctx.Err() will have context.DeadlineExceeded
+	case <-time.After(1 * time.Second):
+		t.Fatalf("Waited too long for DeadlineExceeded ")
+	}
+	w := httptest.NewRecorder()
```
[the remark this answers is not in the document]
chetanvalagit: It difficult to generate error while reading body `buf, err := ioutil.ReadAll(req.Body)` and because of that it will be difficult to create production like behavior.

protoc-gen-twirp/generator.go
```diff
@@ -1334,8 +1334,15 @@ func (t *twirp) generateServerProtobufMethod(service *descriptor.ServiceDescript
...
-	t.P(`    s.writeError(ctx, resp, wrapInternal(err, "failed to read request body"))`)
-	t.P(`    return`)
+	t.P(`  		if ctxErr := context.Canceled; ctxErr == ctx.Err() {`)
+	t.P(`    		s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, ctxErr.Error()))`)
+	t.P(`    		return`)
+	t.P(`  		} else if ctxErr := context.DeadlineExceeded; ctxErr == ctx.Err() {`)
+	t.P(`    		s.writeError(ctx, resp, twirp.NewError(twirp.DeadlineExceeded, ctxErr.Error()))`)
+	t.P(`    		return`)
+	t.P(`  		}`)
+	t.P(`    	s.writeError(ctx, resp, twirp.NewError(twirp.Malformed, "failed to read request body"))`)
```
rhysh, line 1344: This case (returning the "malformed" code when the body read fails) is new since the first round. It's more of a behavior change but I think it's a good one, especially because the server returns that same status if the unmarshal step fails.

rhysh, line 1344: A couple lines down, this function uses the `malformedRequestError` helper to get a similar result in a different case. Please make both/all call sites look the same: this line should use the same helper, or all lines like that one should avoid it. (I don't know exactly why that helper exists.)

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,150 @@ func TestPanicFlushing(t *testing.T) {
...
+		t.Fatalf("Could not even read bytes from response: %q", err.Error())
+	}
+
+	expectedErrMessage := "context deadline exceeded"
+	if !strings.Contains(string(respBytes), expectedErrMessage) {
+		t.Errorf("twirp client err has unexpected message %q, want %q", string(respBytes), expectedErrMessage)
+	}
+}
+
+func TestFailedToReadRequestBodyError(t *testing.T) {
```
rhysh, line 1636: These three test cases have a lot of shared code. It's hard to see from a glance where they differ. Can you factor out the common behavior please?

It also gives a clear opportunity to add additional checks that should be true for all of these cases, including "because these are all client-initiated problems, the http status should be in the range [400,499]".

I find that a `testcase := func(differences) func(t *testing.T) {...}` at the start, followed by a bunch of calls to `t.Run("canceled-context", testcase(args))` is a helpful structure to share some behavior between tests.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,150 @@ func TestPanicFlushing(t *testing.T) {
...
+		}
+		return ctx
+	}}
+	server := NewHaberdasherServer(s, hooks)
+	ctx := context.Background()
+
+	// Make a request, that will call endpoint method
+	req, _ := http.NewRequest(http.MethodPost, "http://testing:8080/twirp/twirp.internal.twirptest.Haberdasher/MakeHat", errReader{})
+	req.Header.Set("Accept", "application/protobuf")
+	req.Header.Set("Content-Type", "application/protobuf")
```
rhysh, line 1654: The tests should also cover behavior of the JSON client.

rhysh (changes requested):

protoc-gen-twirp/generator.go
chetanvalagit: Yes, Original reported came back and ask me for this and shared thread. So thought let me cover that too which I missed at first place.

internal/twirptest/service_test.go
chetanvalagit: I merged three test cases into one. 

We dont have any client and not making any client call. Instead calling server method directly as unit test case. Still wondering how I can check [400, 499] error?

Instead [400, 499] we are checking specific error https://github.com/twitchtv/twirp/pull/294/files#diff-f9d0e6645e9e1b67ccaaceb3da4d91900b84e0cf1ed5d2b7a94f85a9f0423870R1609. Wondering if that is enough?

internal/twirptest/service_test.go
chetanvalagit: Thanks. I think error handling is missing for JSON client also and I end up adding that too. Please let me know if I am wrong.
https://github.com/twitchtv/twirp/pull/294/files#diff-ca331af0419d8cb15dfaf494e87bc6fac163b0f5a897b9927cd1eff3b422b2a8R1272-R1278

.github/workflows/pr-checks.yml
```diff
@@ -29,6 +29,6 @@ jobs:
         import sys
         with open("cover_percentage", "r") as f:
           cov = float(f.readlines()[0])
-          if cov < 12.8:
+          if cov < 12.7:
```
chetanvalagit, line 32: I added code new code, added test case for it in one file (service_test.go). But I think not wise to add test case for each and every test case proto folders to make sure coverage % remains above 12.8.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,127 @@ func TestPanicFlushing(t *testing.T) {
...
 	}
 }
+
+type errReader struct{}
+
+func (errReader) Read(p []byte) (n int, err error) {
+	return 0, fmt.Errorf("test error")
+}
+
+func TestRequestBodyError(t *testing.T) {
```
marioizquierdo, line 1537: Nice update on the tests!

If you can, it would be nice to also add a comment on top of the text to explain a little bit what is trying to accomplish. For example:
```go
// Test failures when reading the request.
// The request could be incomplete because the client timed out, which in most cases is signaled by
// a context cancelation or timeout. In those cases, the twirp handled would properly identify and return those errors.
func TestRequestBodyError(t *testing.T) {
  // ...
}
```

marioizquierdo (changes requested):

protoc-gen-twirp/generator.go
```diff
@@ -1146,6 +1146,20 @@ func (t *twirp) generateServer(file *descriptor.FileDescriptorProto, service *de
...
+	t.P(`func (s *`, servStruct, `) handleRequestBodyError(ctx `, t.pkgs["context"], `.Context, resp `, t.pkgs["http"], `.ResponseWriter, msg string) {`)
+	t.P(`  if ctxErr := context.Canceled; ctxErr == ctx.Err() {`)
+	t.P(`    s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, ctxErr.Error()))`)
+	t.P(`    return`)
+	t.P(`  } else if ctxErr := context.DeadlineExceeded; ctxErr == ctx.Err() {`)
+	t.P(`    s.writeError(ctx, resp, twirp.NewError(twirp.DeadlineExceeded, ctxErr.Error()))`)
+	t.P(`    return`)
+	t.P(`  }`)
+	t.P(`  s.writeError(ctx, resp, malformedRequestError(msg))`)
+	t.P(`}`)
```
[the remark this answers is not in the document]
chetanvalagit: Done

protoc-gen-twirp/generator.go
```diff
@@ -1146,6 +1146,20 @@ func (t *twirp) generateServer(file *descriptor.FileDescriptorProto, service *de
...
 
+	// Write request body read issue
+	t.P(`// handleRequestBodyError is used to handle error when the twirp server cannot read request`)
+	t.P(`func (s *`, servStruct, `) handleRequestBodyError(ctx `, t.pkgs["context"], `.Context, resp `, t.pkgs["http"], `.ResponseWriter, msg string) {`)
+	t.P(`  if ctxErr := context.Canceled; ctxErr == ctx.Err() {`)
+	t.P(`    s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, ctxErr.Error()))`)
+	t.P(`    return`)
+	t.P(`  } else if ctxErr := context.DeadlineExceeded; ctxErr == ctx.Err() {`)
+	t.P(`    s.writeError(ctx, resp, twirp.NewError(twirp.DeadlineExceeded, ctxErr.Error()))`)
+	t.P(`    return`)
```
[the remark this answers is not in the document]
chetanvalagit: Done

internal/twirptest/service_test.go
chetanvalagit: Done

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,131 @@ func TestPanicFlushing(t *testing.T) {
...
+			}
+
+			// validate error message
+			respBytes, err := ioutil.ReadAll(w.Body)
+			if err != nil {
+				t.Fatalf("Could not even read bytes from response: %q", err.Error())
+			}
+
+			expectedErrMessage := tc.errorString
+			fmt.Println(string(respBytes))
```
marioizquierdo, line 1651: Println running wild!

chetanvalagit: Ops good catch. Added for testing.

protoc-gen-twirp/generator.go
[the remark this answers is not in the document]
marioizquierdo: To clarify, I also find the `ctxErr` assignment inside the conditional un-necessarily confusing to read, and it obfuscates the final error message. A developer reading this code may be trying to figure out why they see "context canceled" in the logs, it would be better IMO if we would make that as explicit as possible. `context.Canceled.Error()` is more explicit than `ctxErr.Error()`, specially if that same line means something different on each conditional.

I would prefer if we write this in a more explicit form like this:

```go
func (s *svc2Server) handleRequestBodyError(ctx context.Context, resp http.ResponseWriter, err error, msg string) {
    if ctx.Err() == context.Canceled {
        s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, "failed to read request: context canceled"))
        return
    }
    if ctx.Err() == context.DeadlineExceeded {
        s.writeError(ctx, resp, twirp.NewError(twirp.DeadlineExceeded, "failed to read request: deadline exceeded"))
        return
    }
    s.writeError(ctx, resp, twirp.WrapError(malformedRequestError(msg), err)) // original error can be unwrapped for inspection
}
```
The reason is that this is both easier to read and easier to find when searching for the error message.

I know @chetanvalagit has been going back and forth with this and may not want to keep running in circles, this is more of a question for @rhysh, and while I have a preference here, I don't think this should block the PR any longer, so I'm fine with whatever you both decide now.

protoc-gen-twirp/generator.go
```diff
@@ -1269,8 +1284,8 @@ func (t *twirp) generateServerJSONMethod(service *descriptor.ServiceDescriptorPr
 	t.P(`  reqContent := new(`, t.goTypeName(method.GetInputType()), `)`)
 	t.P(`  unmarshaler := `, t.pkgs["jsonpb"], `.Unmarshaler{AllowUnknownFields: true}`)
 	t.P(`  if err = unmarshaler.Unmarshal(req.Body, reqContent); err != nil {`)
-	t.P(`    s.writeError(ctx, resp, malformedRequestError("the json request could not be decoded"))`)
-	t.P(`    return`)
+	t.P(`       s.handleRequestBodyError(ctx, resp, "the json request could not be decoded", err)`)
+	t.P(`    	return`)
```
marioizquierdo, line 1288: This still has inconsistent indentation. Notice the first line has 3 spaces and the second has 1 tab. The indentation here is maintained manually because it is inside strings, and then is automatically corrected after the code is generated, but it is still useful to indent properly for readability

chetanvalagit: Fixed

protoc-gen-twirp/generator.go
```diff
@@ -1146,6 +1146,21 @@ func (t *twirp) generateServer(file *descriptor.FileDescriptorProto, service *de
 	t.P(`}`)
 	t.P()
 
+	// Write request body read issue
+	t.P(`// handleRequestBodyError is used to handle error when the twirp server cannot read request`)
+	t.P(`func (s *`, servStruct, `) handleRequestBodyError(ctx `, t.pkgs["context"], `.Context, resp `, t.pkgs["http"], `.ResponseWriter, msg string, err error) {`)
+	t.P(`  if ctxErr := context.Canceled; ctxErr == ctx.Err() {`)
+	t.P(`    s.writeError(ctx, resp, twirp.NewError(twirp.Canceled, ctxErr.Error()))`)
```
rhysh, line 1153: It looks like all references to imported packages need to go via the `t.pkgs` map, as they do on the line above. That would hold for the `context` and `twirp` references here and in the next stanza, and for the following `twirp` reference.

(It's unfortunate that we don't have tests to detect that; I don't think it could break based on user input (the name of the protobuf package, for example) since "our" calls to `registerPackageName` come so early, so I'd be surprised if this would be the only part of the generator that skipped the map.)

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,130 @@ func TestPanicFlushing(t *testing.T) {
...
+			} else {
+				defer cancel()
+			}
+			w := httptest.NewRecorder()
+			server.ServeHTTP(w, req)
+
+			// validate error code
+			expectedErrCode := twirp.ServerHTTPStatusFromErrorCode(tc.expectedError)
+			if w.Code != expectedErrCode {
+				t.Errorf("twirp ErrorCode expected to be %q, but found %q", expectedErrCode, w.Code)
```
rhysh, line 1641: The Errorf text isn't right: this compares HTTP status codes.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,130 @@ func TestPanicFlushing(t *testing.T) {
...
+// a context cancelation or timeout. In those cases, the twirp handled would properly identify and return those errors.
+func TestRequestBodyError(t *testing.T) {
+	type contextUpdate func(ctx context.Context) (context.Context, context.CancelFunc)
+	testCases := []struct {
+		testname         string
+		expectedError    twirp.ErrorCode
+		errorString      string
+		reqestType       string
+		contextFunc      contextUpdate
+		isContextCancled bool
```
rhysh, line 1548: If you're making other changes, please fix the spelling here: `isContextCanceled`.

internal/twirptest/service_test.go
```diff
@@ -1527,3 +1527,130 @@ func TestPanicFlushing(t *testing.T) {
...
+// Test failures when reading the request.
+// The request could be incomplete because the client timed out, which in most cases is signaled by
+// a context cancelation or timeout. In those cases, the twirp handled would properly identify and return those errors.
+func TestRequestBodyError(t *testing.T) {
+	type contextUpdate func(ctx context.Context) (context.Context, context.CancelFunc)
+	testCases := []struct {
+		testname         string
+		expectedError    twirp.ErrorCode
+		errorString      string
+		reqestType       string
```
rhysh, line 1546: If you're making other changes, please fix the spelling here: `requestType`.

protoc-gen-twirp/generator.go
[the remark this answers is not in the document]
rhysh: Either is fine, provided the spelling is right. Pulling this out into a specialized function helps with safety, and the tests help with safety. Those two alone are probably sufficient protection against 1/ "cancelled" vs "canceled" or 2/ copy–paste errors that turn "context canceled" into "deadline exceeded".

rhysh (approved):
Thanks for restructuring and expanding the test code. I don't think that any of my remaining comments are deal-breakers. They'd be good to change, but don't need another round of review.

protoc-gen-twirp/generator.go
marioizquierdo: Nice catch! and good point about the tests, thanks Rhysh

chetanvalagit: Great catch :)
Fixed.

marioizquierdo (approved):
LGTM

Issue #222, opened 2020-03-02
Support Protobuf APIV2 


marwan-at-work:
The Go blog just posted that Protobuf's v2 work is officially released: https://blog.golang.org/a-new-go-api-for-protocol-buffers

While I haven't tried it out with Twirp yet, I just wanted to open an issue to track/answer some questions around this potentially breaking change. 

The V1 of go's protobuf library has lived and still does in `github.com/golang/protobuf` while the new version now lives in `google.golang.org/protobuf` -- Therefore I will be referring to those import paths above in this issue to indicate which version I am referring to. 

1. Will Twirp adopting the new protobuf library mean a breaking change for everyone? The blog claims that golang/protobuf v1.4 uses google.golang.org/protobuf under the hood, so maybe that's the first step into migration? 

2. If golang/protobuf v1.4 actually does work with Twirp well, does that mean all Twirp clients must upgrade to v1.4 if the server is using v1.4? 

3. Should Twirp's new major version release (or the one after) incorporate generating the `.twirp.go` file with the new protobuf library?

4. Are any of the new features mentioned in `google.golang.org/protobuf` useful for Twirp in a way that we couldn't do before? Such as printing defaults in json etc.

marwan-at-work:
cc @spenczar

rhysh:
This is a good set of questions. Thanks! Before getting into specifics, to make sure we're on the same page: Twirp generally has a bias towards pragmatism and keeping compatibility. Spencer set the tone in the [announcement post](https://medium.com/twitch-news/twirp-a-sweet-new-rpc-framework-for-go-5f2febbf35f), contrasting it on those two points with gRPC's requirement for http/2 and frequent refactors. Those remain important for Twirp's users (certainly the ones within Twitch).

That sets guidelines for what changes _won't_ be part of Twirp—but what changes _will_? The benefit of the change needs to outweigh the bias for stability. Maybe v2 / "v1.20" has significantly faster (de-)serialization, particularly for binary mode. Maybe v1 gets deprecated and everyone needs to move. Maybe the community moves to v2, and the shims required for v1 compatibility make v1 performance much worse. Those could be enough impetus to move to v2.

1. Yes, application owners specifying v1.4.0 instead of v1.3.x (or older) sounds like a safe first step. This is about what's in the go.mod file of an application, no interaction required from Twirp upstream.

[...]

Issue #301, opened 2021-04-05
Twirp generates code that uses deprecated protobuf libraries


RoarkeRandall:
Our team uses staticcheck, https://staticcheck.io, for linting our go programs. It's convenient to run `staticcheck ./...` on a repo in our ci pipelines. We are getting the following errors for the generated twirp code:

```
pkg/go/service.twirp.go:21:8: package github.com/golang/protobuf/jsonpb is deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead.  (SA1019)
pkg/go/service.twirp.go:22:8: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
```

We are able to ignore these at a file level, but it requires us to manually add a line of code to the top of the generated file. Obviously editing generated code isn't a great solution. 

Is it possible for twirp to update the generated code to use the latest protobuf packages?

mfridman:
The long-term solution is for twirp to update to the new protobuf packages. But, its also a limitation of statickcheck (see this issue and [comment](https://github.com/dominikh/go-tools/issues/429#issuecomment-713267008)).

It takes time for the ecosystem/community to update open-source projects, and statickcheck takes a very all/none approach.

We just ended up with a very unideal situation..

```
checks = [..., "-SA1019"]
```

I know this doesn't help you @RoarkeRandall, but you're not alone.

3ventic:
See also #222.

3ventic:
Twirp v8 generates clients using the new protobuf packages. Refer to the [release](https://github.com/twitchtv/twirp/releases/tag/v8.0.0) for the quick upgrade instructions.

PR #275, opened 2020-09-16
Forced proto3 OPTIONAL feature flag, deps updated 

Solution:
Hello guys,

i've added support for OPTIONAL feature flag which is coming with new version of [protobuf](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/code_generator.h#L108), since the twirp it's not generating the .pb files, it doesn't need to implement anything in code generator, so it'll have no impact on gen-twirp. 

Issue: `.proto is a proto3 file that contains optional fields, but code generator protoc-gen-twirp hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--twirp_out:`

What i did: 
- updated dependencies - it was neccessary for newer version of .pb file of plugin
- add supported feature to plugin response. 

I tried to minimize the impact of upgrading the protobuf deps, but there's lot of changes since the last vendor update.

All tests passing. 

//edit: 
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #304, opened 2021-04-16
Protobuf APIV2

marioizquierdo:
Fixes: #222, #301
Related: #299, #280, #275

# Description

The protobuf library has updated from APIV1 `github.com/golang/protobuf` to APIV2 `google.golang.org/protobuf`. Now Twirp generates code using the new types.

Changes:

 * Generated code no longer depends on `github.com/golang/protobuf/jsonpb`, it now uses `google.golang.org/protobuf/protojson`
 * Generated code no longer depends on `github.com/golang/protobuf/proto`, it now uses `google.golang.org/protobuf/proto`
 * Code generator (`protoc-gen-twirp`) works with the latest version of `google.golang.org/protobuf`
 * Code generator (`protoc-gen-twirp`) improved usage of command line flags. It now accepts `paths=import` (default), `paths=source_relative`, `module={PREFIX}` and works well with multiple `--proto_path` sources. This makes it fully compatible with the flags used on `protoc-gen-go` compiler invocation: https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation
 * Update `go_package` on test proto files to use a fully qualified go import path. This is required by new versions of `protoc-gen-go`.
 * Removed dependency on [gogo-protobuf](https://github.com/gogo/protobuf). Twirp still works with it, but requires a dependency with protobuf APIV1.
 * Removed generated python code on Go tests, that was cluttering the test folders and did not have any effect on them.
 * Simplified Makefile and generate commands, being more explicit makes it easier to know where the binaries are situated.

# Update instructions

 * The runtime library `github.com/twitchtv/twirp` does not have any changes.
 * Generated clients and services (with protoc plugin `github.com/twitchtv/twirp/protoc-gen-twirp`) will update their imports from `github.com/golang/protobuf/proto` (APIV1) to `google.golang.org/protobuf` (APIV2). Projects that make use of the generated clients or service have to update the imports as well.
 * The new `google.golang.org/protobuf` (APIV2) is mostly backwards compatible, but not completely. Some changes that we found needed in our own tests:
   * In the proto file, the `option go_package` is mandatory and must include a "/" (it is supposed to be a full import path). If you have to add a full path to the go_package, you may want to generate with the options `paths=source_relative`. For example: `protoc --go_out=. --go_opt=paths=source_relative --twirp_out=. --twirp_opt=paths=source_relative myfile.proto`
   * Generated message models (structs) now contain a mutex, your linter may complain if the models are copied by value. The solution is to peass pointers instead, or disable the lint for lines where a copy is made on purpose.
   * Check [protobuf go compatibility](https://pkg.go.dev/google.golang.org/protobuf#readme-compatibility) and their [releases](https://github.com/protocolbuffers/protobuf-go/releases) for more details. 


internal/twirptest/importmapping/x/x.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
 package twirp.internal.twirptest.importmapping.x;
-option go_package = "x";
+option go_package = "foo.com/will/be/ignored/x";
```
3ventic, line 4: Let's use example.com or a nonexistent Twitch-owned domain.

internal/twirptest/importmapping/y/y.proto
```diff
@@ -1,5 +1,6 @@
 syntax = "proto3";
 
 package twirp.internal.twirptest.importmapping.y;
+option go_package = "foo.com/will/be/ignored/y";
```
3ventic, line 4: Same here.

3ventic (approved):

marioizquierdo:
This PR will be released on May 3rd (next Monday) as `v8.0.0`.
We are doing some more testing and completing documentation meanwhile.

Issue #300, opened 2021-03-25
Support an optional ErrorCoder interface for Go servers


cespare:
This is a proposal for an improvement to how Twirp handlers return errors. I originally brought this up a couple of years ago [on the Twirp Slack](https://gophers.slack.com/archives/C8T4Y2248/p1569285669003400) in a discussion with @spenczar and @marwan-at-work.

To summarize, I propose that we add the following interface to the `twirp` package. A detailed proposal with background information follows.

```
// ErrorCoder may be implemented by third-party error types returned by
// Twirp handler functions to indicate a Twirp error code to report.
//
// If a non-nil error e returned by a handler is not an Error but is an
// ErrorCoder (or wraps one as indicated by errors.As), the server responds
// with an error using the code given by e.TwirpErrorCode and a message
// given by e.Error.
type ErrorCoder interface {
	TwirpErrorCode() ErrorCode
}
```

If this proposal is acceptable, I'm happy to send a PR.

---

## Background

At my company we use Twirp extensively for making RPCs across systems. We sometimes need to convey particular error conditions (entity not found, say) and handle them differently on the client side. We can do this by sending a particular Twirp error code as the response.

The way this works today is that if a handler returns an error that implements `twirp.Error`, that Twirp error is used for the response. If the handler returns a non-nil error that does not implement `twirp.Error`, then it is considered a generic internal error and wrapped using `twirp.InternalErrorWith`.

This works well enough if the handler makes decisions about different error conditions directly. However, if the error is created a few functions down the call chain, it works less well:

* The function that generates the error might not be Twirp-specific, so it's not appropriate to return a `twirp.Error`, which is very much Twirp-specific. For instance, we have some servers that have shared internal logic which is used for generating web pages as well as Twirp responses and we don't want to generate Twirp errors in non-Twirp routes.
* The error generated down the call stack might be wrapped with other errors before the handler route gets it.

To work around this issue, a common pattern is for our servers to use a function to match errors to various internal error variables/types and return the appropriate Twirp error:

```
func toTwirpError(err error) error {
	switch {
	case errors.Is(err, errJobFinished):
		return twirp.NewError(twirp.FailedPrecondition, err.Error())
	case errors.Is(err, errNoLogDataAfterWait):
		return twirp.NewError(twirp.DeadlineExceeded, err.Error())
	case errors.As(err, new(jobNotFoundError)):
		return twirp.NotFoundError(err.Error())
	// ... several more cases ...
	default:
		return err
	}
}
[...]

PR #323, opened 2021-06-12
Match server twirp errors using errors.As

marioizquierdo:
Issue #300

Allow custom errors to present themselves as `twirp.Errors` using Go 1.11 `errors.As`.


marioizquierdo:
If we proceed with this update, we will need to provide a way for existing services to verify if the update would catch hidden wrapped errors. Here is an error hook that could be used with some instructions: https://github.com/twitchtv/twirp/issues/300#issuecomment-860100481

PR #336, opened 2021-11-04
Don't check the response body close error in the client

cespare:
If we successfully read the response all the way to EOF and decode it, that constitutes a successful RPC. If the body close happens to return a non-nil error, the client shouldn't care.

The context where this came up is that Go 1.17.3 seemingly introduced a bug ([reported here](https://github.com/golang/go/issues/49366)) where sometimes the response body Close method returns a spurious error. This meant that upgrading to Go 1.17.3 caused Twirp requests to start failing for us.

The actual bug is on the Go side, but the fact that nobody noticed and the tests didn't catch it is more evidence that checking this error is uncommon and unnecessary.

(And if I may editorialize slightly: this is why the errcheck linter is not a good blanket policy.)


mellis:
Hello, thank you for the PR. The change has some extraneous changes due to an older version of `protoc` being used. Can you please use the same version that was previously used, `v3.17.1`, to keep it a bit cleaner.

cespare:
> Hello, thank you for the PR. The change has some extraneous changes due to an older version of `protoc` being used. Can you please use the same version that was previously used, `v3.17.1`, to keep it a bit cleaner.

Sure, done. Though if you care about that, ISTM that [the tooling](https://github.com/twitchtv/twirp/blob/218743be42fe3a91a73b4d4fb2d1b6b6a47abefb/check_protoc_version.sh#L20) ought to check for the particular version.

mellis:
Thank you for the update, looks good 👍

> Though if you care about that, ISTM that the tooling ought to check for the particular version.

I agree.

PR #338, opened 2021-11-09
twirp.ErrorCode implements twirp.Error

marioizquierdo:
## What if error codes could also be errors?

This will allow alternative error constructors to the current `twirp.NewError` constructor and similar helpers, to return simpler errors, for example:

```go
return nil, twirp.NotFound

return nil, twirp.NotFound.WithMsg("specific thing was not found")

return nil. twirp.NotFound.WithMsg("foobar").WithMeta("foo", "bar")
```

this is the equivalent of:

```go
return nil, twirp.NewError(twirp.NotFound, "not_found")

return nil, twirp.NewError(twirp.NotFound, "specific thing was not found")

return nil, twirp.NewError(twirp.NotFound, "foobar").WithMeta("foo", "bar")
```

And it will be consistent for every twirp error code.

## Example endpoint

Making `twirp.ErrorCode` to also implement a `twirp.Error` and an `error`, allows a new idiom to construct twirp errors:

```go
func (s *MyService) MyMethod(ctx context.Context, req *mysvc.MyRequest) (*mysvc.MyResponse, error) {
  if req.UserId == "" {
    return nil, twirp.InvalidArgument.WithMsg("user_id is required")
  }
  
  user, err := s.DB.LoadUser(ctx, req.UserId)
  if err != nil {
    return nil, twirp.WrapError(twirp.Internal, err)
  }

  if user == nil {
    return nil, twirp.NotFound
  }

  return &mysvc.MyResponse{}, nil
}
```

For comparison, the same code would look like this without the error code constructors:
Today, this is how returning errors looks like in most Twirp services:

```go
func (s *MyService) MyMethod(ctx context.Context, req *mysvc.MyRequest) (*mysvc.MyResponse, error) {
  if req.UserId == "" {
    return nil, twirp.NewError(twirp.InvalidArgument, "user_id is required")
  }
  
  user, err := s.DB.LoadUser(ctx, req.UserId)
  if err != nil {
    return nil, twirp.WrapError(twirp.InternalError("internal"), err)
  }

  if user == nil {
    return nil, twirp.NotFoundError("not_found")
  }

  return &mysvc.MyResponse{}, nil
}
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -239,6 +239,20 @@ const (
 	NoError ErrorCode = ""
 )
 
+// ErrorCode implements twirp.Error, as a plain error with the same code and msg string values.
+
+func (c ErrorCode) Code() ErrorCode                { return c }
+func (c ErrorCode) Msg() string                    { return string(c) }
+func (c ErrorCode) WithMeta(key, val string) Error { return NewError(c, string(c)).WithMeta(key, val) }
+func (c ErrorCode) Meta(key string) string         { return "" }
+func (c ErrorCode) MetaMap() map[string]string     { return nil }
+func (c ErrorCode) Error() string                  { return string(c) }
+
+// WithMsg returns a copy of the error but with the specified msg.
+func (c ErrorCode) WithMsg(msg string) Error {
+	return NewError(c, msg)
+}
+
 // ServerHTTPStatusFromErrorCode maps a Twirp error type into a similar HTTP
 // response status. It is used by the Twirp server handler to set the HTTP
 // response status code. Returns 0 if the ErrorCode is invalid.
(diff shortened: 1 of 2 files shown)


marioizquierdo:
Thinking a little more about this ... really not a good idea 😅 

There's something bizarre about making the code to also be a twirp error and also be an error. The main issue will be confusion around comparison. Once of the design pillars of Twirp is that the Twirp error returned by the endpoint should be the same that is received by the client. Returning `twirp.NotFound` will be confusing because in the other side you can not just do `twerr == twirp.NotFound`.

I was just exploring ideas, and I think what I wanted was a more consistent way to build errors in the service. After iterating in the idea a little more, I came out with another PR: #339

Closing this PR in favor of #339

PR #341, opened 2021-11-24
Docs: Fix installation commands

liamg:
The installation guide in the docs uses a slightly outdated `go install` command as this now requires a version to install (e.g. `@latest`) as the suffix.

This is a small change to add this `latest` version info.


wmatveyenko:
Thanks for your contribution!

PR #346, opened 2022-01-10
replace map[string]string => map[string]interface{} for error meta

max107:
Replace `Error` meta from `map[string]string` => `map[string]interface{}`

*Description of changes:*

With this MR we can return any struct via meta key in Error struct.


wmatveyenko:
Hi, This looks like a breaking change. Can you share the reasoning behind it?

max107:
> Hi, This looks like a breaking change. Can you share the reasoning behind it?

@wmatveyenko Sometime we should return extra information in error response without marshal it into json as a string, "as is". 

Original code:

```json
{
	"meta": {
		"foo": "bar"
	}
        // or
	"meta": {
		"foo": "{\"bar\":\"baz\"}"
	}
}
```

Current solution

```json
{
	"meta": {
		"foo": {
			"bar": "baz"
		}
	}
}
```

max107:
It's a minimal breaking change, we replace string with interface{}

mfridman:
This is changing an exported interface and the type, so it's more than just a "minimal" breaking change.

Fwiw I'd like to see a change like this, as custom returned errors have always been a pain point. Maybe in a major version bump?

max107:
up

wmatveyenko:
Hi. We discussed this internally and are not going to accept this change. `meta` is not meant for generic JSON serialization, but for key value pairs. Twirp errors are string to string conversions and if you need more complex data returned in the message, you can use Base64 encoded strings for this purpose. The marshaling of map of strings to interface{} could also cause an error, which can introduce additional complexity during error handling.

Thanks!

PR #352, opened 2022-03-29
add-meta-cause-to-wrapped-error

ReyOrtiz:
*Description of changes:*
Adding `cause` meta when wrapping twirp error. It helps when returning the error response when it fails to protojson.unmarshal, it adds the specific error. For example, current error response:
```
{
    "code": "malformed",
    "msg": "the json request could not be decoded"
}
```

With this change:
```
{
    "code": "malformed",
    "msg": "the json request could not be decoded",
    "meta": {
        "cause": "proto: (line 22:17): invalid value for enum type: \"XYZ\""
    }
}
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -92,6 +92,7 @@ func (code ErrorCode) Errorf(msgFmt string, a ...interface{}) Error {
 // The wrapped error can be extracted later with (github.com/pkg/errors).Unwrap
 // or errors.Is from the standard errors package on Go 1.13+.
 func WrapError(twerr Error, err error) Error {
+	twerr = twerr.WithMeta("cause", err.Error())
 	return &wrappedErr{
 		wrapper: twerr,
 		cause:   err,


wmatveyenko:
Hi, we discussed this internally and would like to know more about the problem you are solving with this change.
Thanks!

ReyOrtiz:
Hello, thanks for replying. We have several rpc services generated from .proto files and some endpoints are exposed to external callers. If the clients calling the endpoints have an invalid payload they only get the `malformed` error without any message indicating what the issue is. This change would help the external callers to troubleshoot and fix any payload error without requiring assistance as the error response would have enough information.

I tried using Interceptors and Server Hooks but since it's an error the interceptor would not be called and with server hooks we can't modify the error response body.

ReyOrtiz:
Hello @wmatveyenko, is there any update about this change? Thanks!

ReyOrtiz:
Hi @wmatveyenko, sorry to reach out again. Is there any news about this PR? Thanks.

wmatveyenko:
Sorry for the wait here. We were discussing this internally. We don't think making this change is the right way to accomplish this, since it opens up information about errors that application owners may not want to be returned to callers. This is a problem that pluggable encoders may be able to fix, but that requires a larger amount of thinking and development effort. 

Here is a possible solution that you can do for yourself that can give you this functionality ahead of time: 
You could write a response writer that usually forwards to the underlying one, but when it gets the error callback, you could unplug the underlying writer and marshal the error response yourself. This would most likely involve the implementation of a Twirp server hook, an http.handler middleware and a ResponseWriter wrapper.

PR #355, opened 2022-05-14
fix ./example/ to be in sync with ./docs/example.md

ra1u:
./example is modified in a way to be same as docs/example.md

There are some modifications to docs/example.md , particularly to command  protoc that was incompatible with recent versions of protoc.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #361, opened 2022-07-25
Revert "fix ./example/ to be in sync with ./docs/example.md"

wmatveyenko:
Reverts twitchtv/twirp#355 because it causes the tests to fail.


rhysh (approved):
LGTM, thanks. Weird that the CI didn't run against #355 (or at least, didn't report status to that pull request).

PR #363, opened 2022-08-04
Replace references to Error type with Error code

davidblnc:
Just noticed looking at the docs that according to the routing section errors have a "type" field, but everywhere else this field is called "code" (I guess it was "type" in a previous version).


wmatveyenko:
Thanks for your contribution.

PR #367, opened 2022-09-16
Remove deprecated ioutil from generator

marwan-at-work:
The ioutil std lib package has been deprecated for a few releases now, this PR removes all references to them.


marioizquierdo:
What is the latest version of Go that would not work with this change? I seems that `io.ReadAll` was introduced in Go 1.16. Is that right? The current version is 1.19 and Twirp promises to keep compatibility with the last 2 versions, so removing support for Go 1.15 and below seems fine. Can we confirm this?

marwan-at-work:
@marioizquierdo Yes, the docs on the ioutil package say 1.16 as well https://pkg.go.dev/io/ioutil

wmatveyenko:
Thank for your contribution. Can you add - 
> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

marwan-at-work:

wmatveyenko:
Thanks for your contribution.

marwan-at-work:
@wmatveyenko thank you (and to the others) for keeping the project maintained. Let me know if you plan on cutting a patch release. Thanks!

PR #379, opened 2023-05-04
Add larking to list of twrip runtimes

emcfarlane:
Hello, I’ve been working on a service to help use gRPC transcoding. The implicit methods map really neatly over to twirps protocol. In testing it looks good. Error support was added and works but by translating grpc status to twirp errors it doesn’t support all the meta fields and two error codes. Please see test case here: https://github.com/emcfarlane/larking/blob/main/benchmarks/twirp_test.go 

Add  https://github.com/emcfarlane/larking as a twirp server runtime.


swiftyspiffy:
Hello! Apologies for the delay on this.

This is a neat project, but I think for the time being we're not currently adding new personal projects. We'll probably develop a criteria for project submissions at some point, but we've got nothing at the moment.

emcfarlane:
@swiftyspiffy no worries! Thanks for getting back to me.

Issue #390, opened 2024-03-12
Documentation website domain triggers Chrome's warning 


riprsa:
<img width="673" alt="image" src="https://github.com/twitchtv/twirp/assets/44776391/76a18aaf-3f09-406f-831f-78cfbdd6ad20">

Docs website, of course, is: https://twitchtv.github.io/twirp/

I hope it is possible to contact Google from your side and fix this 😅

swiftyspiffy:
We're aware of this issue and tracking internally. No progress yet, but definitely known.

swiftyspiffy:
https://issuetracker.google.com/issues/342242650
```
Thank you for submitting your submission. We have successfully fetched and processed your digital asset link manifests across the following domains:
  * twitch.tv
  * twitchtv.github.io

Within a few days, lookalike warnings should stop showing on these sites. If you're still seeing the warnings in 1 week, feel free to follow-up here.
```

Will leave this open until issue is confirmed gone.

swiftyspiffy:
Confirmed the issue is no longer present. Closing.

PR #393, opened 2024-05-18
add .well-known to static directory to resolve automated lookalike warning

swiftyspiffy:
*Issue:* #390 

*Description of changes:*

Following google's instructions here: https://chromium.googlesource.com/chromium/src/+/master/docs/security/lookalikes/lookalike-domains.md#Removing-Lookalike-Warnings-from-a-site

I know it says not to use subdomains, but I figure because the error itself only presents on the specific subdomain, that maybe we can get away with only specifying the subdomain. We can update this later to github.io if needed.

The same change needs to be made against twitch.tv, which I'll work to have done internally.


howsleya (approved):
lgtm

